PR #25, opened 2018-12-31
Add class for showing pop-up warnings to user

AlexejheroYTB:


QModManager/ErrorMessage.cs
```diff
@@ -0,0 +1,22 @@
...
+namespace QModManager
+{
+    public static class ErrorMessage
+    {
+        public static void ShowError(string error, Action<bool> function, string leftButtonText = "Yes", string rightButtonText = "No")
+        {
+            // Still need to implement leftButtonText and rightButtonText
+            uGUI_SceneConfirmation confirmation = uGUI.main.confirmation;
+
+            if (leftButtonText == "" || leftButtonText == null) confirmation.no.gameObject.SetActive(false);
```
PrimeSonic: Use `string.IsNullOrEmpty(` when `string` values for null and empty.

AlexejheroYTB: Addressed in commit 837a2bd

AlexejheroYTB:
Still need to figure out how to change the text - It's hard when you don't have a scene debugger

ahk1221:
I'll take a look.

AlexejheroYTB:
There is an issue with the buttons not working properly - this pr will be put on hold until #26 is fixed

QModManager/Error.cs
```diff
@@ -0,0 +1,37 @@
+﻿using System;
+using System.Diagnostics;
+using System.IO;
+using UnityEngine.UI;
+
+namespace QModManager
+{
+    internal static class Error
+    {
+        internal static void ShowError(string error, Action onLeftButtonPressed = null, Action onRightButtonPressed = null, string leftButtonText = "See Log", string rightButtonText = "Close")
```
PrimeSonic: Optional parameters of the same type can make for messy calls of the method in question.  
You may often find that you will have to constantly add the parameter names to make this work.

**Proposal:**  
Create a new abstract class `ButtonAction` that exposes a `string Text` property and a `Action ButtonPressed` property.  
Then, you can create `LeftButtonAction : ButtonAction` and `RightButtonAction : ButtonAction` classes and use them as optional parameters here with `null` as their default.  
```CSharp
ShowError(string error, LeftButtonAction leftButton = null, RightButtonAction rightButton = null);
```
This way, there is no ambiguity in the call and no need to use parameter names to clearly express which text and which action belong to which button.

QModManager/Error.cs
```diff
@@ -0,0 +1,37 @@
...
+            else confirmation.yes.gameObject.GetComponentInChildren<Text>().text = leftButtonText;
+            if (string.IsNullOrEmpty(rightButtonText)) confirmation.no.gameObject.SetActive(false);
+            else confirmation.no.gameObject.GetComponentInChildren<Text>().text = rightButtonText;
+
+            confirmation.Show(error, delegate (bool leftButtonClicked)
+            {
+                if (leftButtonClicked) onLeftButtonPressed.Invoke();
+                else onRightButtonPressed.Invoke();
+
+                Console.Write("");
```
PrimeSonic: If there is a reason for this, at least say so in a comment.

QModManager/Error.cs
```diff
@@ -0,0 +1,37 @@
+﻿using System;
+using System.Diagnostics;
+using System.IO;
+using UnityEngine.UI;
+
+namespace QModManager
+{
+    internal static class Error
```
PrimeSonic: Best to choose a more descriptive class name. This one could cause confusion.  
Suggestion: `internal static class ErrorDialogs`

QModManager/Error.cs
AlexejheroYTB: The buttons have a stupid tendency to not work 90% of the time. I made some changes that broke them but I wasn't what caused it. This line is a leftover from testing. I will remove it

QModManager/Error.cs
AlexejheroYTB: I personally don't see a point in bundling two parameters in a class, as 3 out of the 4 optional parameters will never be used in the release buld - only for testing

QModManager/Error.cs
AlexejheroYTB: Addressed in commit 10ef4b8

QModManager/ErrorDialog.cs
```diff
@@ -0,0 +1,67 @@
...
+using System.Diagnostics;
+using System.IO;
+using UnityEngine;
+using UnityEngine.UI;
+
+namespace QModManager
+{
+    internal static class Dialog
+    {
+        internal static void Show(string error, Action onLeftButtonPressed = null, Action onRightButtonPressed = null, string leftButtonText = "See Log", string rightButtonText = "Close", bool blue = false)
```
ahk1221: Maybe have the text parameters before the Action parameters.

AlexejheroYTB: I don't think that this is necessary as there will always be `Action` parameters when there are the `string` parameters. Also, again, most of them won't be used in the final build. And if they will, they'll be named.

ahk1221: Guess it makes sense.

PR #28, opened 2019-01-03
Uninstall by removing patch call to QMM instead of using a dll backup

ahk1221:


QModManager/Injector.cs
```diff
@@ -93,27 +91,35 @@ internal void Remove()
...
+                    if(instruction.OpCode == OpCodes.Call && instruction.Operand.ToString().Equals("System.Void QModInstaller.QModPatcher::Patch()"))
+                    {
+                        patchMethodCall = instruction;
+                    }
                 }
 
-                Console.WriteLine();
-                Console.WriteLine("Cannot uninstall, file 'Assembly-CSharp-qoriginal.dll' is missing");
-                Console.WriteLine("To uninstall, you will need to verify game contents ons steam");
+                awakeMethod.Body.GetILProcessor().Remove(patchMethodCall);
```
AlexejheroYTB: Maybe a `null` check for `patchMethodCall`?

ahk1221: Oh yeah, forgot to check for that.

QModManager/Injector.cs
```diff
@@ -93,27 +91,36 @@ internal void Remove()
...
+                    {
+                        patchMethodCall = instruction;
+                    }
                 }
 
-                Console.WriteLine();
-                Console.WriteLine("Cannot uninstall, file 'Assembly-CSharp-qoriginal.dll' is missing");
-                Console.WriteLine("To uninstall, you will need to verify game contents ons steam");
+                if(patchMethodCall != null)
+                    awakeMethod.Body.GetILProcessor().Remove(patchMethodCall);
```
AlexejheroYTB: If its not null, you need to tell the user that the IsInjected check succeeded, but the patch method call wasn't found. (`An unexpected error has occurred.` and exit with exit code `3`)

ahk1221: There is no way for that to happen considering the IsInjected check uses the same condition that I use to get the call instruction.

ahk1221: So the null check wasn't even required, its just "there".

AlexejheroYTB: Even if its not required, its better to have it

PR #29, opened 2019-01-03
Version check

AlexejheroYTB:
## `VersionCheck.cs`
- A new file which checks for a new version from the `version.json` file in this repository on the `master` branch
- Uses async web requests
- If a newer version is available, shows the user a blue dialog
- If there are errored mods, the dialog isn't shown

## Notes
- ~~I will handle merge conflicts myself~~
- ~~I will need to change the branch from where `version.json` is loaded from `version-check` to `master` once the pr is merged~~
- ~~The text on the buttons is too big for the `Download` text to show properly. this will be fixed in a later commit~~

image


QModManager/Patcher.cs
```diff
@@ -24,20 +25,22 @@ internal static void Patch()
...
                 }
                 patched = true;
 
                 Hooks.Patch();
                 StartLoadingMods();
-                Hooks.Update += ShowErroredMods;
+                Hooks.Update += ShowDialog;
 
-                Hooks.OnLoadEnd();
+                VersionCheck.Check();
```
PrimeSonic: Because this is now reaching out over the internet from a user's computer, we should respect the user's decision as to whether or not they wish to allow QMM to do this.

The installer wizard need to have an option somewhere reading something like "Allow QModManager to check for updates?"

That setting will need to be stored and checked here so that this is only executed if the user has given QMM permission to do so.

AlexejheroYTB: Added information about this in the installer, and a toggle options in the mods menu to disable it.

QModManager/Injector.cs
```diff
@@ -93,8 +94,7 @@ internal void Remove()
 
                 // Remove backup file if it exists
                 string backupFilePath = Path.Combine(managedDirectory, "Assembly-CSharp.qoriginal.dll");
-                if (File.Exists(backupFilePath))
-                    File.Delete(backupFilePath);
+                if (File.Exists(backupFilePath)) File.Delete(backupFilePath);
```
PrimeSonic: Unnecessary change.  
Even single line if-block statements are easier read when on their own line and indented.

QModManager/VersionCheck.cs
```diff
@@ -0,0 +1,105 @@
...
+                UnityEngine.Debug.Log("Could not get latest version!");
+                return;
+            }
+            Version version = new Version(versionStr);
+            if (version == null)
+            {
+                UnityEngine.Debug.Log("Could not get latest version!");
+                return;
+            }
+            if (!version.Equals(QMod.QModManagerVersion) && QModPatcher.erroredMods.Count <= 0) Dialog.Show($"There is a newer version of QModManager available: {version.ToString()}  (current version: {QMod.QModManagerVersion.ToString()})",
```
PrimeSonic: Again, this is pretty hard to read, even with full width available.  
Please keep single statement code blocks on their own line and nested under their parent if-block.

QModManager/VersionCheck.cs
```diff
@@ -0,0 +1,105 @@
...
+            if (version == null)
+            {
+                UnityEngine.Debug.Log("Could not get latest version!");
+                return;
+            }
+            if (!version.Equals(QMod.QModManagerVersion) && QModPatcher.erroredMods.Count <= 0) Dialog.Show($"There is a newer version of QModManager available: {version.ToString()}  (current version: {QMod.QModManagerVersion.ToString()})", 
+                () => Process.Start(nexusmodsURL), leftButtonText: "Download", blue: true);
+        }
+
+        internal const string VersionURL = "https://raw.githubusercontent.com/QModManager/QModManager/version-check/latest-version.txt";
```
PrimeSonic: Any change that this could be pulled out into an external config file instead of being hard-coded into the source?

AlexejheroYTB: The file will always be the same, and will not change. I don't see any reason to put it in an external config file

PrimeSonic: Unless you control GitHub itself, there is no long term guarantee that user content will always be handled like this.  
Put a pin on this idea for a future revision.

QModManager/VersionCheck.cs
AlexejheroYTB: Addressed in commit 35f3801

QModManager/Injector.cs
AlexejheroYTB: Addressed in commit 35f3801

AlexejheroYTB:
Unless anyone wants to add anything, I shall merge this pr tomorrow.

PR #30, opened 2019-01-03
🚧 Create language helper

AlexejheroYTB:


QModManager/LanguageHelper.cs
```diff
@@ -0,0 +1,89 @@
...
+            if (strings.ContainsKey(key)) return Location.Main;
+            if (backupStrings.ContainsKey(key)) return Location.Backup;
+            return Location.Missing;
+        }
+        public static string Get(string key)
+        {
+            if (string.IsNullOrEmpty(key)) return null;
+            if (strings.ContainsKey(key)) return strings[key];
+            if (backupStrings.ContainsKey(key)) return backupStrings[key];
+            return null;
```
PrimeSonic: If this is text that is going to be printed on screen,  
it may be best to return the `key` to be used as the value instead of `null`.

This way, even in the event of an error, there will still be some intelligible text.

AlexejheroYTB: Agreed. Will change in the next commit

AlexejheroYTB:
This pr will be merged into #29, not into the `1.4-dev` branch

AlexejheroYTB:
I will also change the key names in English.json to be more friendly and use them through the project.

AlexejheroYTB:
Currently, `Executable` has no way of knowing the language, since it not run with subnautica. Until I find a good way to handle this I will make it use English

AlexejheroYTB:
Also - the project is now live on poeditor so that anyone can contribute: https://poeditor.com/join/project/P4Z17VDP6L

PrimeSonic:
> Currently, `Executable` has no way of knowing the language, since it not run with subnautica. Until I find a good way to handle this I will make it use English

Perhaps a simple drop-down as part of the first screen of the installer would be enough.  
You can either try to detect Steam language selection, or you can have the user select a language in that first window, with "English" being already selected should then simply want to hit "Next".

QModManager/LanguageHelper.cs
AlexejheroYTB: Addressed in commit 664d768

PrimeSonic:
Not sure how much you have left but this is looking pretty good.

AlexejheroYTB:
Closing for now. Might reopen later

PR #33, opened 2019-01-06
🚧 Small logger

AlexejheroYTB:
Might add more methods


AlexejheroYTB:
This code should be in SMLHelper.

PR #42, opened 2019-01-30
Below zero compatibility. Also some more Mac stuff

AlexejheroYTB:


AlexejheroYTB:
This is just a first concept, please review very harshly!

AlexejheroYTB:
Currently, the stuff to get the game and os are in the executable. I might move that into the dll, since it will be required by it too.

AlexejheroYTB:
There will need to be some cleanup for logs and such. (next pr)

QModManager/Patcher.cs
```diff
@@ -266,7 +260,75 @@ internal static bool LoadMod(QMod mod)
...
+            {
+                if (mod.Game == game) return true;
+
+                if (!nonApplicableMods.Contains(mod)) nonApplicableMods.Add(mod);
+                if (!erroredMods.Contains(mod)) erroredMods.Add(mod);
+                return false;
+
+            }).ToList();
+
+            Console.WriteLine($"\nQMOD ERROR: The following {GetOtherGame()} mods were not loaded because {game.ToString()} was detected:");
```
PrimeSonic: Log as a Warning, not an Error.

Executable/Program.cs
```diff
@@ -196,6 +173,55 @@ internal static void Main(string[] args)
             }
         }
 
+        internal static void GetInfo(out OS os, out string windowsDirectory, out string macDirectory)
```
PrimeSonic: `os` is an `out` parameter for this method, but I'm not seeing it get assigned anywhere.

QModManager/Patcher.cs
AlexejheroYTB: If there are any errors with LoadOrder/Dependencies, it's logged as an error

PrimeSonic: Semantics. The mod loader here correctly identifies that the mod shouldn't be loaded.  
It didn't encounter an error while trying to load it.  
And since it won't be loaded, it shouldn't cause issues for the player.

Hence why I would call this a warning. There are actions the user could take, but ultimately, things can proceed as normal.

Executable/Program.cs
AlexejheroYTB: I must've deleted this by accident. Addressed in commit 0dcc903

QModManager/Patcher.cs
AlexejheroYTB: Addressed in commit fc9d40a

Executable/Program.cs
```diff
@@ -46,67 +54,36 @@ internal static void Main(string[] args)
                     Environment.Exit(1);
                 }
 
-                string windowsDirectory = Path.Combine(Environment.CurrentDirectory, "../..");
-                string macDirectory = Path.Combine(Environment.CurrentDirectory, "../../../../..");
+                GetInfo(out os, out string windowsDirectory, out string macDirectory);
```
ahk1221: Why have the directory variable be separate for each OS? Wouldn't it be fine if it was just a single variable?

AlexejheroYTB: That is true, and it will also cut down on some if elses

AlexejheroYTB: Addressed in commit 779c377

PR #43, opened 2019-01-31
Rework for Below Zero(Only)

kohanis:
In Below Zero half of the code was changed, all but 1 mods that I checked were down . So what's the point of trying to make a comparable version of the loader? It's built on the latest versions of Game Assembly and Unity dlls
It's not for merge with this branch, but u haven't 2.0)


AlexejheroYTB:
We are already working on a Below Zero rework for QModManager, so that the **same** installer and files will work with both games: #42

kohanis:
> We are already working on a Below Zero rework for QModManager, so that the **same** installer and files will work with both games: #42

yeah, i seen it, but it's still use old unity dlls and old Assembly. They really recoded a lot. But ur choose, ok

AlexejheroYTB:
The dlls should still work even if they are older versions. We only use the unity dlls to compile, we don't distribute them. Unity never removes features without deprecating  them first, so it should still work. We'll have someone test it later

PR #44, opened 2019-02-01
Code cleanup for 1.4

AlexejheroYTB:
~~This should be merged after #42.~~


QModManager/QMod.cs
```diff
@@ -40,15 +40,38 @@ internal static QMod FromJsonFile(string file)
...
-                Console.WriteLine(e.ToString());
+                Logger.Error($"\"mod.json\" deserialization failed for file \"{file}\"!");
+                Debug.LogException(e);
 
                 return null;
             }
         }
     }
+    
+    internal class FakeQMod : QMod
```
ahk1221: This can be replaced by a static method that returns a normal QMod but with the updated values.

AlexejheroYTB: Addressed in commit 60f9a4b

PR #46, opened 2019-02-02
Below zero compatible installer

AlexejheroYTB:
The same installer file which works both for SN and BZ


AlexejheroYTB:
Installer video soon

AlexejheroYTB:
https://youtu.be/enG95fz0sLQ
Also shows the installation process and proves it actually works

PR #50, opened 2019-02-09
Basic piracy checking

AlexejheroYTB:
Unfortunately, as I haven't been able to fix the audio, the behavior will just be QMM not loading, no main menu music, no cursor, and inactive buttons. 

If I can manage to fix the audio, I will make the video play instead, but I don't want to delay the release of 1.4


AlexejheroYTB:
Seeing as there have been a lot of features introduced, the next version of QModManager will be 2.0
With that being said, I would love to find a way to enable custom sounds before the release of 2.0, this way we don't have to postpone it.

PR #54, opened 2019-02-14
Check for duplicate mod ids

AlexejheroYTB:


QModManager/Patcher.cs
```diff
@@ -346,6 +348,35 @@ private static bool LoadMod(QMod mod)
...
+        private static void RemoveDuplicateModIDs()
+        {
+            List<QMod> duplicateModIDs = new List<QMod>();
+
+            foreach (QMod mod in sortedMods)
+            {
+                List<QMod> matchingMods = sortedMods.Where((QMod qmod) => qmod.Id == mod.Id).ToList();
+                if (matchingMods.Count > 1)
+                {
+                    matchingMods.Do((QMod qmod) =>
```
PrimeSonic: Would it be possible to write this without LINQ?  
I'm not even seeing where anything is being added to `duplicateModIDs`.

AlexejheroYTB: Line 362 is the line you're looking for.  
Also `Do` is not from LINQ, it's from Harmony. (literally)
  
If you want, removing `Do` and using normal C# can be easily achieved.

PrimeSonic: Yeah. For something this critical, I don't mind writing extra lines of code as long as the end result is clear and readable.  
Feel free to use `foreach`. I just want to be able to see exactly when a duplicate is detected and how.

AlexejheroYTB: Addressed in commit 58f17ff

PR #55, opened 2019-02-14
A bunch of random things that shouldn't really be in the same pull request, but relate to eachother

AlexejheroYTB:
- Enable Unity audio using AssetsTools.NET
- Move Injector.cs to Executable
- Add ☠️ video + audio
- Update C# version to 7.3


AlexejheroYTB:
Again, I have managed to mess up the base branch. Will fix after #54 is merged

AlexejheroYTB:
YES! IT WORKS! FINALLY! YES!

Executable/Injector.cs
```diff
@@ -10,6 +10,7 @@ internal class Injector
     {
         private readonly string gameDirectory;
         private readonly string managedDirectory;
+        private readonly string globalgamemanagers;
```
ahk1221: Should be `globalGameManagers`

AlexejheroYTB: I named the variable lowercase because the actual file name is lowercase too.

QModManager/PirateCheck.cs
```diff
@@ -34,46 +34,59 @@ private void Start()
...
-            }*/
+            }
 
             private void Update()
             {
                 RuntimeManager.GetBus("bus:/master").setMute(true);
                 UWE.Utils.alwaysLockCursor = true;
                 UWE.Utils.lockCursor = true;
-                SceneManager.GetActiveScene().GetRootGameObjects().Do(go =>
+                string[] bannedGOs =
```
PrimeSonic: For how this is being used, make this `static readonly HashSet<string> BannedGameObjectNames` instead.

QModManager/PirateCheck.cs
```diff
@@ -162,28 +175,74 @@ private IEnumerator PlayVideo()
...
+                                {
+                                    html += ch;
+                                }
+                            }
+                        }
+                    }
+                }
+                catch
+                {
+                    return "";
```
PrimeSonic: `string.Empty`

QModManager/PirateCheck.cs
```diff
@@ -162,28 +175,74 @@ private IEnumerator PlayVideo()
...
+            }
+            private static string GetHtmlFromUri(string resource)
+            {
+                string html = string.Empty;
+                HttpWebRequest req = (HttpWebRequest)WebRequest.Create(resource);
+                try
+                {
+                    using (HttpWebResponse resp = (HttpWebResponse)req.GetResponse())
+                    {
+                        bool isSuccess = (int)resp.StatusCode < 299 && (int)resp.StatusCode >= 200;
```
PrimeSonic: Don't cast to `int`.  
Instead use the proper `HttpStatusCode` enum value
https://docs.microsoft.com/en-us/dotnet/api/system.net.httpstatuscode?view=netframework-4.7.2

QModManager/PirateCheck.cs
```diff
@@ -162,28 +175,74 @@ private IEnumerator PlayVideo()
...
                 yield return StartCoroutine(PlayVideo());
-            }*/
+            }
+
+            private static bool CheckConnection(string hostedURL = "http://www.google.com")
+            {
+                try
+                {
+                    string HtmlText = GetHtmlFromUri(hostedURL);
+                    if (HtmlText == "")
```
PrimeSonic: `string.IsNullOrEmpty`

Executable/AudioFixer.cs
```diff
@@ -0,0 +1,37 @@
+﻿using AssetsTools.NET;
+using AssetsTools.NET.Extra;
+using System.Collections.Generic;
+using System.IO;
+
+namespace QModManager
+{
+    internal static class AudioFixer
+    {
+        internal static void ChangeDisableUnityAudio(string path, bool newValue)
```
PrimeSonic: Someone is going to forget what `newValue` is supposed to mean.  
Give this parameter a more meaningful name.

QModManager/PirateCheck.cs
```diff
@@ -110,7 +123,7 @@ private void ShowText()
             {
                 DestroyImmediate(gameObject.GetComponent<RawImage>());
                 Text text = gameObject.AddComponent<Text>();
-                text.text = $"Oopsie poopsie, an error has occured!\nQModManager couldn't be initialized.\nTo fix this issue, turn on your internet connection,\nthen restart {(Patcher.game == Patcher.Game.Subnautica ? "Subnautica" : "Below Zero")}.\n(Is the game pirated?)";
+                text.text = $"An error has occured!\nQModManager couldn't be initialized.\nPlease turn on your internet connection,\nthen restart {(Patcher.game == Patcher.Game.Subnautica ? "Subnautica" : "Below Zero")}.";
```
PrimeSonic: Don't use `\n` for line breaks in text.  
Use `Environment.Newline` instead which is safer and will always use the correct line break characters for the current OS.

Executable/AudioFixer.cs
AlexejheroYTB: The parameter literally means the new value of the `Disable Unity Audio` setting

QModManager/PirateCheck.cs
AlexejheroYTB: So far, `string.Empty` has never been used in this project. I know that there's a small difference between it and `""`, the latter creating an object, but I don't feel like it's necessary to change that.

QModManager/PirateCheck.cs
AlexejheroYTB: Not casting to an int would mean having to manually check for 5 to 7 values that can all represent success.

PrimeSonic: You can still use `<` and `>=` with int based enums.

QModManager/PirateCheck.cs
AlexejheroYTB: Addressed in commit 58e5345

QModManager/PirateCheck.cs
AlexejheroYTB: Addressed in commit 58e5345

QModManager/PirateCheck.cs
AlexejheroYTB: Unless i'm missing something, the operators can only be used on the enums if they are casted to an integer beforehand

PrimeSonic: It's probably a newer C# 7 feature. But it's perfectly valid and keeps arbitrary codes more readable

QModManager/PirateCheck.cs
AlexejheroYTB: `\n` is used in soo many places now, and I think that there isn't an actual need to change it. Windows uses `\r\n` for line breaks _for backward compatibility_, and most other operating systems use `\n`. Despite that, the `\n` character behaves like `\r\n` on Windows operating systems, meaning that there is no actual need to change it. (IN MY OPINION)

QModManager/PirateCheck.cs
PrimeSonic: Just trying to get you to pick up some good habits `^_~`

QModManager/PirateCheck.cs
ahk1221: Wasn't there some sort of incompatibility because of using `\n` on other operation systems? I remember there was. But in any case, Environment.NewLine is definitely better.

QModManager/PirateCheck.cs
AlexejheroYTB: Addressed in commit 5e0fb15

QModManager/PirateCheck.cs
AlexejheroYTB: `\r\n` is only used for Unix based operating systems, such as Windows.
`\n` is used by all other operating systems, including Mac.

Quote from [here](https://stackoverflow.com/questions/1015766/difference-between-n-and-environment-newline):
> Note that when writing to the console, `Environment.NewLine` is not strictly necessary. The console stream will translate `\n` to the appropriate new-line sequence, if necessary.

`Environment.NewLine` just seems messy to me, creating very long lines or strings that span across multiple lines. So far there haven't been any errors with `\n` on _any_ operating systems, but if there are, I will definitely change it.

QModManager/PirateCheck.cs
AlexejheroYTB: Addressed in commit 18f3005

PR #63, opened 2019-03-02
Output folder structure to the log

AlexejheroYTB:


AlexejheroYTB:
This can be improved _a lot_, by looping through the files once and storing their information. I will handle this in an upcoming commit

QModManager/Utility/IOUtilities.cs
```diff
@@ -0,0 +1,180 @@
...
+                    {
+                        size /= 1024;
+                        if (size < 1024)
+                        {
+                            unit = "GB";
+                        }
+                        else
+                        {
+                            size /= 1024;
+                            unit = "TB";
```
PrimeSonic:
```CSharp
var sizes = new[] { "B", "KB", "MB", "GB", "TB" };
int unit = 0;
while (size > 1024)
{
    unit ++;
    size /= 1024;
};
...
return number + sizes[unit];

File: QModManager/Utility/IOUtilities.cs [hunk 2]
```diff
@@ -0,0 +1,180 @@
...
+                        }
+                    }
+                }
+            }
+
+            string number = size.ToTwoDecimalString();
```
PrimeSonic: string.Format("F2");  
https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-numeric-format-strings

QModManager/Utility/IOUtilities.cs
```diff
@@ -0,0 +1,180 @@
...
+            return number + unit;
+        }
+        private static long CalculateFolderSizeRecursively(string directory)
+        {
+            try
+            {
+                long size = 0;
+                foreach (string dir in Directory.GetDirectories(directory))
+                {
+                    size += CalculateFolderSizeRecursively(dir);
```
PrimeSonic: The total size of folders seems irrelevant for trouble-shooting.  
Why not just show only the size of files and simply this?

QModManager/Utility/IOUtilities.cs
```diff
@@ -0,0 +1,180 @@
...
+
+        private static string GenerateSpaces(int spaces)
+        {
+            string s = "";
+            for (int i = 1; i <= spaces; i+=4)
+                s += "|   ";
+            return s;
+        }
+
+        private static int GetFileCountRecursively(string directory)
```
PrimeSonic: File counts also seems irrelevant for what this could be used for.  
Is it really needed?

QModManager/Utility/IOUtilities.cs
AlexejheroYTB: Or, as @PPVVDD suggested, `Math.Log`

QModManager/Utility/IOUtilities.cs
AlexejheroYTB: Addressed in commit c2d45a3

QModManager/Utility/IOUtilities.cs
AlexejheroYTB: Not really, I just think it's "fancy"

QModManager/Utility/IOUtilities.cs
AlexejheroYTB: Addressed in commit 2bc60d6

QModManager/Utility/IOUtilities.cs
AlexejheroYTB: This extension method is from `Assembly-CSharp-firstpass`, I don't think there's any harm in using it.

AlexejheroYTB: After looking at the method in dnSpy, this is exactly what it does. Changed in commit 1a86b08

dnSpy screenshot of ToTwoDecimalString

PR #72, opened 2019-04-13
Added information concerning Linux users

Eonfge:
I've not tested it extensively, but your current version of QModManager works and some mods, notably Map, work. I hope that with this extra info, future users can easily use mods on Linux.


PrimeSonic:
After the various updates to support Unix style file paths on Mac, I don't expect there to be any issues with Linux using Steam Proton.

Eonfge:
@PrimeSonic That was very fast. Was writing a more detailed issue explaining as-to-why this would be a nice addition, but thanks for the quick merge.

PR #74, opened 2019-04-13
QModManager public API

AlexejheroYTB:


PrimeSonic:
Looking good so far.

AlexejheroYTB:
What other methods should I add?

PrimeSonic:
> What other methods should I add?

Sorry. I can't offer any ideas for what more to add to this.

PR #76, opened 2019-04-16
Ability to specify version ranges of dependencies

AlexejheroYTB:
This is done via a new mod.json field: `VersionDependencies` of type `Dictionary<string, string>`.

Parsing is done with [semver.net](https://github.com/adamreeve/semver.net), and it also accepts ranges (more info [here](https://github.com/adamreeve/semver.net/blob/master/README.md))


README.md
```diff
@@ -109,6 +115,10 @@ Example: `"BestMod.QMod.Patch"`
   "Author": "Awesome Guy",
   "Version": "1.0.0",
   "Dependencies": [ "DependencyModID" ],
+  "VersionDependencies": { 
+    "VeryImportantMod": "1.2.x", 
+    "AnotherVeryImportantMod": "2.1.2 - 2.3.2"
```
PrimeSonic: The samples provided here should showcase the most common usage of the feature.  
In most cases, dependencies are "_this specific version or later_".  
Make the example something like that.

PR #77, opened 2019-04-18
QModAPI: Dependency injection, ability to specifiy QModManager as a VersionDependency, .xml documentation

AlexejheroYTB:


QModManager/Messages.cs
```diff
@@ -0,0 +1,71 @@
+﻿using System;
+using System.Linq;
+using System.Reflection;
+
+namespace QModManager
+{
+    public static partial class QModAPI
+    {
+        [AttributeUsage(AttributeTargets.Method, AllowMultiple = false, Inherited = false)]
+        public class MessageReceiverAttribute : Attribute
```
PrimeSonic: Are nested classes necessary?  
Could this all be moved into a `QModManager.API` namespace instead?

QModManager/Messages.cs
```diff
@@ -0,0 +1,71 @@
...
+            }
+        }
+
+        [AttributeUsage(AttributeTargets.Method, AllowMultiple = false, Inherited = false)]
+        public class GlobalMessageRecieverAttribute : Attribute
+        {
+            public GlobalMessageRecieverAttribute() { }
+        }
+
+        public static void SendMessage(QMod mod, string message, params object[] data)
```
PrimeSonic: This is going to need an example mod to showcase how to use this function,  
both on part of the sender and especially on the part of the receiver.

QModManager/Messages.cs
```diff
@@ -0,0 +1,71 @@
...
+            QMod caller = GetMod(Assembly.GetCallingAssembly(), true);
+
+            mod.LoadedAssembly.GetTypes()
+                .SelectMany(t => t.GetMethods())
+                .Where(m => m.IsStatic)
+                .Where(m => m.GetParameters().Length == 3)
+                .Where(m => m.GetCustomAttributes(typeof(MessageReceiverAttribute), false)
+                    .Where(a => ((MessageReceiverAttribute)a).Sender == caller)
+                    .ToArray().Length > 0)
+                .ForEach(m => m.Invoke(null, new object[] { caller, message, data }));
```
PrimeSonic: Can these methods not be detected and cached during the loading process?  
Having to find them all over again each time seems like an unnecessary performance hit.

If we already have a collection of all mod assemblies, why not expand the `QMod` class with cached references to the methods this needs to invoke?

QModManager/QModAPI.cs
```diff
@@ -1,10 +1,11 @@
 ﻿using System.Collections.Generic;
+using System.Collections.ObjectModel;
 using System.Reflection;
 using System.Text.RegularExpressions;
 
 namespace QModManager
 {
-    public static class QModAPI
+    public static partial class QModAPI
```
PrimeSonic: Will the API support dependency injection?

QModManager/API/Messages.cs
```diff
@@ -0,0 +1,65 @@
...
+    public static partial class QModAPI
+    {
+        public static void SendMessage(QMod mod, string message, params object[] data)
+        {
+            if (mod == null) return;
+            if (data == null) data = new object[] { };
+
+            QMod caller = GetMod(Assembly.GetCallingAssembly(), true);
+
+            if (mod.MessageReceivers.TryGetValue(caller, out List<MethodInfo> methods))
```
PrimeSonic: Is the `MessageReceivers` field in the `QMod` class  guaranteed to be not-null by this point?  
Same concern on `BroadcastMessage` method here.

QModManager/API/Messages.cs
```diff
@@ -0,0 +1,65 @@
...
+
+        public static void BroadcastMessage(string message, params object[] data)
+        {
+            if (data == null) data = new object[] { };
+
+            QMod caller = GetMod(Assembly.GetCallingAssembly(), true);
+
+            if (GetAllMods().SelectMany(m => m.MessageReceivers)
+                            .ToDictionary(k => k.Key, v => v.Value)
+                            .TryGetValue(null, out List<MethodInfo> methods))
```
PrimeSonic: If the methods are identified by `message` string, it looks like we could cache which mods implement that message.

When the mods are being loaded and the `MessageReceivers` are being cached, we could also build up data structure that quickly maps a method `message` identifier with the mods that implement that message:  
```CSharp
internal readonly static Dictionary<string, List<QMod>> CachedMessageReceivers = new Dictionary<string, List<QMod>>();
```
This would allow for less searching in the `BroadcastMessage` and `SendMessage` methods.
```CSharp
if (CachedMessageReceivers.TryGetValue(message, out List<QMod> mods))
{
    foreach (QMod mod in mods)
        foreach (MethodInfo method in mod.MessageReceivers)
            method.Invoke(null, new object[] { caller, message, data });
}
```
Making sure this feature doesn't introduce performance penalties will encourage modders to use it.

QModManager/Patcher.cs
```diff
@@ -340,6 +342,23 @@ internal static void RemoveDuplicateModIDs()
...
+                           .Where(m => m.GetCustomAttributes(typeof(MessageReceiver), false).Length == 1)
+                           .GroupBy(m => ((MessageReceiver)m.GetCustomAttributes(typeof(MessageReceiver), false)[0]).Sender)
+                           .Select(g => new KeyValuePair<QMod, List<MethodInfo>>(g.Key, g.ToList()))
+                           .Add(new KeyValuePair<QMod, List<MethodInfo>>(null,
+                   assembly.GetTypes()
+                           .SelectMany(t => t.GetMethods())
+                           .Where(m => m.IsStatic && m.GetParameters().Length == 3)
+                           .Where(m => m.GetCustomAttributes(typeof(GlobalMessageReciever), false).Length == 1)
+                           .ToList()))
+                           .ToDictionary(k => k.Key, v => v.Value);
```
PrimeSonic: All these constraints make sense, but they also look like they could be greatly simplified by the use of an `interface`.  
You wouldn't be able to use `static` methods for this anymore, and modders may need to implement their own singleton patterns for it, but it would make it very easy to inform modders what the correct method signature is for these features.

This is only a suggestion in case you haven't thought of it yet.

QModManager/API/QModAPI.cs
```diff
@@ -0,0 +1,50 @@
...
+        {
+            if (includeErrored)
+                return Patcher.foundMods.AsReadOnly();
+            else if (includeUnloaded)
+                return Patcher.sortedMods.AsReadOnly();
+            else
+                return Patcher.loadedMods.AsReadOnly();
+        }
+
+        public static QMod GetMod(Assembly modAssembly = null, bool includeUnloaded = false, bool includeErrored = false)
```
PrimeSonic: Recommendation:  
Add a separate method for `GetMyMod()` that returns the calling assembly's QMod instance.  
After that, the `GetMod` can always require a `modAssembly` parameter.

QModManager/API/QModAPI.cs
```diff
@@ -0,0 +1,50 @@
...
+
+            foreach (QMod mod in GetAllMods(includeUnloaded, includeErrored))
+                if (mod.LoadedAssembly == modAssembly) return mod;
+            return null;
+        }
+        public static QMod GetMod(string id, bool includeUnloaded = false, bool includeErrored = false)
+        {
+            if (string.IsNullOrEmpty(id)) return null;
+            foreach (QMod mod in GetAllMods(includeUnloaded, includeErrored))
+                if (mod.Id == Regex.Replace(id, "[^0-9a-z_]", "", RegexOptions.IgnoreCase)) return mod;
```
PrimeSonic: If this regex is used in more than one place, the regex itself could be refactored into a single, common location.  
This would avoid duplication and avoid problems in the future if we ever have to update the pattern.

QModManager/API/Messages.cs
AlexejheroYTB: No, it could actually be null... A solution would be to either discard messages that were sent before the mod was loaded or cache them and send them once it has loaded

QModManager/API/Messages.cs
AlexejheroYTB: Methods are not identified by the `message` string, it is used to identify what kind of message is sent and how the `data` object will be used. It might as well have been a part of the `data` array.

QModManager/Patcher.cs
AlexejheroYTB: I dont think this would work well, since the attribute would still be needed to define the sender. Of course, this can by bypassed by sending it in the constructor, but I'm not sure if that would be the best approach

QModManager/API/Messages.cs
AlexejheroYTB: Which one of these do you think would be the way to go?

QModManager/Messages.cs
AlexejheroYTB: Maybe something for the wiki...?

QModManager/API/Messages.cs
AlexejheroYTB: With commit 83306b0, I implemented a check for null `MessageReceivers` in the `QMod` class.

QModManager/Patcher.cs
AlexejheroYTB: Commit 2c4fcc2610bc408: replaced attributes with abstract classes which enforce the constraints

QModManager/API/Messages.cs
```diff
@@ -0,0 +1,83 @@
...
+
+namespace QModManager.API
+{
+    public abstract class MessageReceiver
+    {
+        public IQMod From;
+
+        public MessageReceiver(IQMod from) => From = from ?? throw new ArgumentNullException("from", "The provided mod is null!");
+        public MessageReceiver(Assembly fromAssembly) => From = QModAPI.GetMod(fromAssembly ?? throw new ArgumentNullException(nameof(fromAssembly), "The provided assembly is null!"), true, true) ?? throw new ArgumentException("The provided assembly is not a mod assembly!", nameof(fromAssembly));
+        public MessageReceiver(string fromID) => From = QModAPI.GetMod(fromID ?? throw new ArgumentNullException(nameof(fromID), "The provided ID is null!"), true, true) ?? throw new ArgumentException("No mod matching the provided ID was found!");
```
AlexejheroYTB: Should we not throw an exception if the ID is not found?

AlexejheroYTB: Should we not throw an exception if the ID is not found?

QModManager/API/QModAPI.cs
```diff
@@ -0,0 +1,98 @@
...
+                if (mod.LoadedAssembly == modAssembly) return mod;
+            return null;
+        }
+        IQMod IQModAPI.GetMod(string id, bool includeUnloaded = false, bool includeErrored = false)
+        {
+            if (string.IsNullOrEmpty(id)) return null;
+
+            foreach (QMod mod in GetAllMods(includeUnloaded, includeErrored))
+                if (mod.Id == Regex.Replace(id, Patcher.IDRegex, "", RegexOptions.IgnoreCase)) return mod;
+            return null;
```
PrimeSonic: This is starting to getting really hard to read.
```CSharp
// Use braces if the nested statement occupies more than 1 line
foreach (QMod mod in GetAllMods(includeUnloaded, includeErrored))
{
    if (mod.Id == Regex.Replace(id, Patcher.IDRegex, "", RegexOptions.IgnoreCase)) 
        return mod; // Keep `return` statements on their own line
} 
// Add an empty line after the end of a nested instruction
return null;
```

QModManager/API/Messages.cs
AlexejheroYTB: If the ID is not found, it may be because the mod is missing. In that case, I'm not entirely sure an exception should be thrown

QModManager/API/QModAPI.cs
```diff
@@ -0,0 +1,98 @@
...
+        public static bool ModPresent(string id, bool includeUnloaded = false, bool includeErrored = false)
+            => Main.ModPresent(id, includeUnloaded, includeErrored);
+
+        #region Non-static
+
+#pragma warning disable CS1066 // The default value specified will have no effect because it applies to a member that is used in contexts that do not allow optional arguments
+
+        void IQModAPI.MarkAsErrored(Assembly modAssembly = null)
+        {
+            if (ErroredMods.Contains(modAssembly)) return;
```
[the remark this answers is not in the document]
AlexejheroYTB: The warning is about something else: https://stackoverflow.com/questions/5683111/warning-from-explicitly-implementing-an-interface-with-optional-parameters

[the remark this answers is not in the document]
PrimeSonic: Oh, I just noticed that this is a method from an interface (I was too sleepy still).

Yes, absolutely _do not_ include optional parameters in interfaces. Instead, make a separate method for each variation you want to have.  
I did this for SMLHelper ([example](https://github.com/SMLHelper/SMLHelper/blob/DevMar2019/SMLHelper/Interfaces/ISpriteHandler.cs)), and while tedious, it's necessary.

QModManager/API/QModAPI.cs
```diff
@@ -0,0 +1,103 @@
...
+        public static bool ModPresent(Assembly modAssembly, bool includeUnloaded = false, bool includeErrored = false)
+            => Main.ModPresent(modAssembly, includeUnloaded, includeErrored);
+        public static bool ModPresent(string id, bool includeUnloaded = false, bool includeErrored = false)
+            => Main.ModPresent(id, includeUnloaded, includeErrored);
+
+        #region Non-static
+
+        void IQModAPI.MarkAsErrored(Assembly modAssembly = null)
+        {
+            modAssembly = modAssembly ?? Assembly.GetCallingAssembly();
```
AlexejheroYTB: `Assembly.GetCallingAssembly()` should be replaced with an utility method which gets the calling assembly by stacktrace

PrimeSonic: Pinch from [SMLHelper](https://github.com/SMLHelper/SMLHelper/blob/DevMar2019/SMLHelper/Utility/ReflectionHelper.cs#L173)

AlexejheroYTB: I know, I was leaving a comment for myself 😅

QModManager/API/Messages.cs
```diff
@@ -0,0 +1,123 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+
+namespace QModManager.API
```
AlexejheroYTB: Some docs would be nice. Maybe in an .xml file

AlexejheroYTB:
I've ran into a lot of issues with the `MessageReceiver`s, I will need to change how the target QMod is defined (maybe by a property instead of in the constructor)

PR #81, opened 2019-05-11
Use Doorstop for injection

AlexejheroYTB:
Use [Doorstop](https://github.com/NeighTools/UnityDoorstop/) for injection instead of manual Mono injection
### Pros
- Works after game updates
- There will be no cases where QModManager is installed but it's not patched, or it's patched but not installed
- No more file replacing, no more IO exceptions
### Cons
- Users cannot verify game files to uninstall QModManager
- No mac support
- What are we going to do with the [AudioFixer](../blob/dev/master/Executable/AudioFixer.cs)?


AlexejheroYTB:
@legojoshua12 @Eonfge If this pull request will be merged, there will be no mac/linux support, and it would need to be handled differently.

Eonfge:
@AlexejheroYTB A few things come to my mind:

First off, I don't know it this actually breaks Linux compatibility. After all, the Linux version runs the game in Wine (Proton), which is best described as a kind of container that converts Windows system calls to Linux. It might well be that this new dll is fully covered by this system, in which case Linux support will be unaffected. 

Mac on the other hand, will be left in the cold. Unless you want to publish different versions of the same .dll (.elf & .so in unix terms), they will have no way to run QModManager

Second, and this is a more fundamental question to you: Is there any reason to change the existing back-end so drastically? 

Following SemVar standards, this would become QModManager 3.0, since you're intentionally breaking backwards compatibility. Doing so is always an option, but I'm not sure if this is really required considering that modding Subnautica is in its infancy.

What I would propose, is to keep 2.0.2 around as a legacy (Mac compatible) version, and to publish 3.0 as a newer version. At that point, mod developers can choose themselves if they prefer cross platform compatibility or if they prefer new features. This way, you can see how the market develops. Perhaps Mac and Linux support is more important to most mod makers then state-of-the-art features.

AlexejheroYTB:
@Eonfge 
> First off, I don't know it this actually breaks Linux compatibility. After all, the Linux version runs the game in Wine (Proton), which is best described as a kind of container that converts Windows system calls to Linux. It might well be that this new dll is fully covered by this system, in which case Linux support will be unaffected.

According to the author of Doorstop, there is also a linux variant of this tool, so a linux version could work, but it would require extra work.

image

> Mac on the other hand, will be left in the cold. Unless you want to publish different versions of the same .dll (.elf & .so in unix terms), they will have no way to run QModManager

That, is what I am afraid of. I know that the old patching method could just be kept, but it would mean that we would have to have two different versions of QModManager, one for windows and one for mac, with only the mac one containing the injector

> Second, and this is a more fundamental question to you: Is there any reason to change the existing back-end so drastically?

Nothing with QModManager itself has been changed, only with the way it is injected. This way of injection is better than what we used before, which would manually inject the IL code using mono.

> Following SemVar standards, this would become QModManager 3.0, since you're intentionally breaking backwards compatibility. Doing so is always an option, but I'm not sure if this is really required considering that modding Subnautica is in its infancy.

I am not sure what you mean. Backwards compatibility is not broken.

>What I would propose, is to keep 2.0.2 around as a legacy (Mac compatible) version, and to publish 3.0 as a newer version. At that point, mod developers can choose themselves if they prefer cross platform compatibility or if they prefer new features. This way, you can see how the market develops. Perhaps Mac and Linux support is more important to most mod makers then state-of-the-art features.

I would like to hear @legojoshua12's opinion on that

Eonfge:
Uhm, Doorstap does not need to be compiled for Linux when using it in Wine (Proton). You'll use the Windows installer with the Windows .dll files, which will then be translated to Linux. 

The fact that @ghorsington says that it can be compiled to linux gives hope: That means that there are no Windows exclusive elements in the code, which means that Wine should have it easy. In this case, Linux support might actually be saver then Mac support. After all, Linux runs the Windows version, while Mac won't be able to do that. Still, no guaranties for the time being. 

If you have a beta build, let me know, I'll give it a testrun with Subnautica.

AlexejheroYTB:
About mac support:
image

denikson:
> According to the author of Doorstop, there is also a linux variant of this tool, so a linux version could work, but it would require extra work.

From what I managed to read, you already know of using Wine/Proton to run Doorstop. This is actually the way some other modding communities currently do it (i.e. Risk of Rain 2).

If you require native Linux implementation you mentioned, it can be found at [NeighTools/UnityDoorstop.Linux](https://github.com/NeighTools/UnityDoorstop.Linux). As noted, no prebuilt binaries yet since it's mainly a PoC. 

> The fact that @denikson says that it can be compiled to linux gives hope: That means that there are no Windows exclusive elements in the code

As far as core functionality of Doorstop goes, there is indeed nothing specific to Windows. But then comes all the junk. The reason why Doorstop and Doorstop.Linux are separate is because Windows requires quite a bit of different quirks to get the game to load a DLL of one's choice. A pain, yes. However as a consequence I can keep Linux version just to about ~200 LOC.

> About mac support:

As for macOS support, it should be possible to edit Doorstop.Linux, since macOS and Linux share a lot of components it uses. It's just guesswork based on some research, though. As I did indeed say, I don't have an ability to port it right now. If you're interested in Doorstop (and I thank you if you do 😄), maybe someone else has the time and know-how?

Dependencies/doorstop_config.ini
```diff
@@ -0,0 +1,3 @@
+[UnityDoorstop]
+enabled=true
+targetAssembly=QModManager/QModInstaller.dll
```
PrimeSonic: If this means that players can't simply verify the integrity of their game files to uninstall QMM,  
then how are they going to uninstall or disable it at all?

What about people who switch between Nitrox and QMM?  
What will it look like when a player wants to disable all mods for one reason or another?

AlexejheroYTB: Manual uninstallation from add & remove programs or this: 

Screenshot_2019-05-09-09-45-32

AlexejheroYTB: Uninstalling QModManager by verifying game files was never intentional, it was just a side effect of how the injection was done. 
A file verification would cause QModManager to appear installed in the programs list, but not patched.

legojoshua12:
Just from looking at this, this biggest problem to me is the uninstalling problem. Set aside the loss of mac support, making something unfamiliar to users seems like a bad idea. Just looking at this, is there really a true reason to change much of the core function of QMM? There really isn't a whole lot left for QMM to do, that has become much to mod devs. Subnautica isn't going to receive many, if any, updates in the future, so why do we need a tool for that? I think we should most certainly keep the current stuff we have going.  It also sounds like there could be some problems with linux, which means what this drastic change is going to do is create more headaches than solve.

As for mac loss, that also is a big issue. Is something really an upgrade if it ever removes compatibility? This large scale conversion would require someone to fork the project and keep the separate versions supported, which is going to ultimately lead to conflict for mod developers and issues for users down the line.

As pointed out by @PrimeSonic, this is going to make problems for Nitrox users as well.  Making it harder to uninstall QMM is not a plus, so that means we would HAVE to make something work for disabling Nitrox and using QMM or vise versa.  That is a huge undertaking so its completely up to you how you want to make a monumental shift in development here, but from my input I see it as impractical and has significantly more drawbacks than it does for pluses.

AlexejheroYTB:
> Just from looking at this, this biggest problem to me is the uninstalling problem. [...] As pointed out by @PrimeSonic, this is going to make problems for Nitrox users as well. Making it harder to uninstall QMM is not a plus, so that means we would HAVE to make something work for disabling Nitrox and using QMM or vise versa.

Personally, I fail to see what "uninstalling problem" you are talking about. Many, if not most of the people who use QModManager remove it either by uninstalling it from the programs list or by running `QModManager.exe` and uninjecting. The only people who are most likely to uninstall QModManager by verifying game files are QModManager developers or people that don't know how to actually uninstall it. Even though it is counter-intuitive, it is much better than manual file injection, with which we have a high chance of getting errors (`IOExceptions` for files in use, the notorious `BadImageFormatException`, etc)

> Subnautica isn't going to receive many, if any, updates in the future

Subnautica will update to Unity 2018.2 very soon, which will break most, if not all, mods. I want to get all backwards incompatible features done my then. (Including the SMLHelper merge)

> It also sounds like there could be some problems with linux, which means what this drastic change is going to do is create more headaches than solve. As for mac loss, that also is a big issue. Is something really an upgrade if it ever removes compatibility? This large scale conversion would require someone to fork the project and keep the separate versions supported, which is going to ultimately lead to conflict for mod developers and issues for users down the line.

Something like the current `QModManager.exe` could be kept specifically for mac (and maybe linux) users, and the injection process could be changed only for Windows (and maybe linux)

> That is a huge undertaking so its completely up to you how you want to make a monumental shift in development here

I do not want to make a decision solely by myself, as one of the original devs, I think @RandyKnapp's opinion should be heard.

AlexejheroYTB:
@Eonfge Do you have a discord account? I'd like to discuss about this matter in a... more friendly environment

Eonfge:
@AlexejheroYTB I don't have a Discord, and I'm not sure what you wish to discuss in more detail. 

I'm not too concerned about future linux support, especially since Denikson mentioned a known fix. All I'll have to do is update the readme with a dll. override when QModManager 2.1 comes out.

> Subnautica will update to Unity 2018.2 very soon, which will break most, if not all, mods. I want to get all backwards incompatible features done my then. (Including the SMLHelper merge)

What actually informed you that Subnautica will get an engine update in the future? I mean, it's cool that the studio behind it keeps supporting the game, but this comes a bit out of nowhere. Or are you referring to Subnautica: Below Zero?

> I do not want to make a decision solely by myself, as one of the original devs, I think RandyKnapp's opinion should be heard.

It also comes down a bit to time and effort. Is it worth it for you? I have easy talking from the sideline, as I'm actually in no way involved with the coding. You should judge for yourself if this is worth the time investment. If Subnautica (Below Zero) gets many more updates in the coming years, then better to use the injector to save yourself many future headaches. Perhaps Mac OS X support is just not worth it. Shame for those who have the game on Mac OS X, but so be it. At this point, I think that both choices are valid, and well researched. 

Who knows, perhaps there is help out there. Anybody reading this who doesn't mind building QModManager for Mac OS X?

AlexejheroYTB:
> What actually informed you that Subnautica will get an engine update in the future? I mean, it's cool that the studio behind it keeps supporting the game, but this comes a bit out of nowhere. Or are you referring to Subnautica: Below Zero?

One of the @UnknownWorlds developers confirmed that Subnautica will be updated to a newer version of Unity (5.6.2 > 2018.2) a while ago, and actually the current [experimental](https://subnautica.fandom.com/wiki/Experimental_Mode) build of Subnautica is using Unity 2018.2. It's only a matter of time until the engine update is pushed to live.

image

> It also comes down a bit to time and effort. Is it worth it for you? I have easy talking from the sideline, as I'm actually in no way involved with the coding. You should judge for yourself if this is worth the time investment. If Subnautica (Below Zero) gets many more updates in the coming years, then better to use the injector to save yourself many future headaches. Perhaps Mac OS X support is just not worth it. Shame for those who have the game on Mac OS X, but so be it. At this point, I think that both choices are valid, and well researched. Who knows, perhaps there is help out there. Anybody reading this who doesn't mind building QModManager for Mac OS X?

Mac support has been requested a very long time ago (#5), and the first working version was created by @Zebralear (#15). Since then, @legojoshua12 has worked on an actual mac `.pkg` installer which was supposed to be released with `2.0`, but was delayed due to some issues. We _have_ mac support, and personally, I would rather we not lose it with this update.

AlexejheroYTB:
Also, currently, the [AudioFixer](https://github.com/QModManager/QModManager/blob/dev/master/Executable/AudioFixer.cs) injects/uninjects when QModManager is installed/uninstalled. I believe that can be changed so it is done as startup/shutdown

AlexejheroYTB:
After we successfully merge @SMLHelper with QModManager, I will start working on this pull request. I genuinely think that using Doorstop would be a better form of injection. As long as the linux version can work perfectly with the windows version of Doorstop, there is no need to make a separate linux build. As for mac, the executable could be kept as an alternative form of injection... at least if we don't have any better ideas.

Executable/AudioFixer.cs
```diff
@@ -12,16 +12,20 @@ internal static void ChangeDisableUnityAudio(string path, bool newValue, Patcher
         {
```
AlexejheroYTB: The `AudioFixer` will have to be moved from the `Executable` project to the `QModManager` project

Dependencies/doorstop_config.ini
AlexejheroYTB: Paths will have to be updated in both `QModManager` and `SMLHelper` to match the new folder locations

Installer/QModsInstallerScript.iss
```diff
@@ -51,49 +51,54 @@ CloseApplications=False
...
+Source: "..\Dependencies\AssetsTools.NET.dll"; DestDir: "{app}\QModManager"; Flags: IgnoreVersion
+Source: "..\Dependencies\cldb.dat"; DestDir: "{app}\QModManager"; Flags: IgnoreVersion
+Source: "..\Dependencies\cldb2018.dat"; DestDir: "{app}\QModManager"; Flags: IgnoreVersion
+Source: "..\Dependencies\doorstop_config.ini"; DestDir: "{app}\QModManager"; DestName: "doorstop_config.ini.doorstop"; Flags: IgnoreVersion
+Source: "..\Dependencies\Mono.Cecil.dll"; DestDir: "{app}\QModManager"; Flags: IgnoreVersion
+Source: "..\Dependencies\SemVer.dll"; DestDir: "{app}\QModManager"; Flags: IgnoreVersion
+Source: "..\Dependencies\SMLHelper.dll"; DestDir: "{app}\QModManager"; Flags: IgnoreVersion
+Source: "..\Build\QModInstaller.dll"; DestDir: "{app}\QModManager"; Flags: IgnoreVersion
+Source: "..\Build\QModInstaller.xml"; DestDir: "{app}\QModManager"; Flags: IgnoreVersion
+Source: "..\Build\QModManager.exe"; DestDir: "{app}\QModManager"; Flags: IgnoreVersion
```
AlexejheroYTB: We could put QModManager.exe in the directly in the game directory instead of the `QModManager` subdirectory so that users can see it and run it as an alternative form of patching/unpatching

AlexejheroYTB:
I am putting this in the "Optional" milestone as it's not important and I don't think it will reach 3.0

Executable/Injector.cs
```diff
@@ -1,25 +1,15 @@
-﻿using Mono.Cecil;
-using Mono.Cecil.Cil;
-using System;
+﻿using System;
 using System.IO;
-using System.Linq;
 
 namespace QModManager.Executable
 {
     internal class Injector
```
AlexejheroYTB: @PrimeSonic we will need to talk about the injector and how we plan on moving things forward with this

PrimeSonic: At this point, knowing that Mac and PC are using different installers anyways, I am not opposed to  the idea of using DoorStop for the injector on PC.  
If we can include a simple toggle to disable this on demand without having to go through an un-installation, that would be even better.

AlexejheroYTB: That's what I was thinking: Having the injector located in the base game folder acting as a toggle for QModManager which basically copies/deletes the doorstop files from the game's folder.

Although I'm not sure if I want to have the mac executable and windows executable different. @legojoshua12 what would you prefer?

AlexejheroYTB: Either that, or we could add a new entry in the steam library (and a desktop shortcut for epic games users) to run subnautica modded.

legojoshua12: I actually don't think having different installers is bad since its already that way, unless we can figure out how to combine the two into one installer.  Since it uses manual mono injection before, does that mean with two separate installers we would have one doorstop and the other on mono still?

AlexejheroYTB:
There will be a ton of merge conflicts after #103 is merged, so I will probably re-write this pull request from scratch on the new branch.

AlexejheroYTB:
Postponed. Will be added in another version

PR #93, opened 2019-05-24
Merge with SMLHelper

AlexejheroYTB:
Merge QModManager with @SMLHelper


Examples/mod.json
```diff
@@ -0,0 +1,9 @@
+{
+  "Id": "SMLHelperExampleMod",
+  "DisplayName": "Example Mod For SMLHelper",
+  "Author": "ahk1221 & PrimeSonic & AlexejheroYTB",
+  "Version": "Updated for the latest version of SMLHelper",
+  "Enable": true,
+  "AssemblyName": "Example mod.dll",
+  "EntryMethod": "SMLHelper.V2.Examples.ExampleMod.Patch"
```
AlexejheroYTB: Needs to be updated

QModManager/API/SMLHelper/Assets/CustomFishPrefab.cs
```diff
@@ -0,0 +1,123 @@
+﻿namespace QModManager.API.SMLHelper.Assets
+{
+    using UnityEngine;
+    using QModManager.API.SMLHelper.Utility;
+    using Logger = QModManager.API.SMLHelper.Logger;
```
AlexejheroYTB: Move usings before namespaces or keep them this way?

QModManager/API/SMLHelper/Crafting/TechData.cs
```diff
@@ -0,0 +1,136 @@
...
+            return TechType.None;
+        }
+    }
+
+    /// <summary>
+    /// A class for representing a required ingredient in a recipe.
+    /// </summary>
+    /// <seealso cref="IIngredient" />
+    /// <seealso cref="TechData"/>
+    public class Ingredient : IIngredient
```
AlexejheroYTB: Some extension methods or implicit conversions could be added here

QModManager/API/SMLHelper/Handlers/CustomFishHandler.cs
```diff
@@ -0,0 +1,51 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+    using QModManager.API.SMLHelper.Assets;
+    using System.IO;
+
+    /// <summary>
+    /// Class to manage registering of fish into the game
+    /// </summary>
+    public static class CustomFishHandler
```
AlexejheroYTB: Dependency injection?

QModManager/API/SMLHelper/Handlers/CustomFishHandler.cs
```diff
@@ -0,0 +1,51 @@
...
+        public static List<TechType> fishTechTypes = new List<TechType>();
+
+        /// <summary>
+        /// Registers a CustomFish object into the game
+        /// </summary>
+        /// <param name="fish">The CustomFish that you are registering</param>
+        /// <returns>The TechType created using the info from your CustomFish object</returns>
+        public static TechType RegisterFish(CustomFish fish)
+        {
+            Logger.Log($"[FishFramework] Creating fish: {fish.displayName}");
```
AlexejheroYTB: Update this to match SMLHelper standards

QModManager/API/SMLHelper/Handlers/CustomFishHandler.cs
```diff
@@ -0,0 +1,51 @@
...
+                swimSpeed = fish.swimSpeed,
+                swimRadius = fish.swimRadius,
+                swimInterval = fish.swimInterval,
+                pickupable = fish.isPickupable,
+                isWaterCreature = fish.isWaterCreature
+            };
+
+            if(!string.IsNullOrEmpty(fish.spriteFileName))
+            {
+                SpriteHandler.RegisterSprite(type, Path.Combine(Environment.CurrentDirectory + "/QMods/",fish.spriteFileName));
```
AlexejheroYTB: Use Path.Combine inside of Path.Combine instead of concatenating the strings

QModManager/API/SMLHelper/Initializer.cs
```diff
@@ -0,0 +1,51 @@
...
+    using System;
+    using System.Reflection;
+    using Harmony;
+    using Patchers;
+
+    public class Initializer
+    {
+        private static HarmonyInstance harmony;
+
+        public static void Patch()
```
AlexejheroYTB: Need to add QMod-SML integration

QModManager/API/SMLHelper/Patchers/BioReactorPatcher.cs
```diff
@@ -0,0 +1,19 @@
+﻿namespace QModManager.API.SMLHelper.Patchers
+{
+    using Harmony;
+    using System.Collections.Generic;
+
+    internal class BioReactorPatcher
```
AlexejheroYTB: All patchers can be turned static

QModManager/API/SMLHelper/Assets/CustomFishPrefab.cs
AlexejheroYTB: @PrimeSonic I'd like to know your opinion

QModManager/API/SMLHelper/Assets/CustomFishPrefab.cs
```diff
@@ -0,0 +1,123 @@
...
+            }
+            Locomotion loco = mainObj.GetOrAddComponent<Locomotion>();
+            loco.useRigidbody = rb;
+            mainObj.GetOrAddComponent<EcoTarget>().type = EcoTargetType.Peeper;
+            mainObj.GetOrAddComponent<CreatureUtils>();
+            mainObj.GetOrAddComponent<VFXSchoolFishRepulsor>();
+            SplineFollowing spline = mainObj.GetOrAddComponent<SplineFollowing>();
+            spline.locomotion = loco;
+            spline.levelOfDetail = mainObj.GetOrAddComponent<BehaviourLOD>();
+            spline.GoTo(mainObj.transform.position + mainObj.transform.forward, mainObj.transform.forward, 5f);
```
AlexejheroYTB: Very cluttered

QModManager/API/SMLHelper/Assets/CustomFishPrefab.cs
```diff
@@ -0,0 +1,123 @@
...
+            behaviour.splineFollowing = spline;
+
+            if (pickupable)
+            {
+                Logger.Log("[FishFramework] Adding pickupable component", LogLevel.Debug);
```
AlexejheroYTB: Could be removed

QModManager/API/SMLHelper/Handlers/CustomFishHandler.cs
```diff
@@ -0,0 +1,66 @@
...
+    using QModManager.API.SMLHelper.Assets;
+    using System.IO;
+    using QModManager.API.SMLHelper.Interfaces;
+
+    /// <summary>
+    /// Class to manage registering of fish into the game
+    /// </summary>
+    public class CustomFishHandler : ICustomFishHandler
+    {
+        public static ICustomFishHandler Main { get; } = new CustomFishHandler();
```
AlexejheroYTB: Some public members don't have XML comments.

QModManager/API/SMLHelper/Handlers/KnownTechHandler.cs
```diff
@@ -0,0 +1,171 @@
...
+    using UnityEngine;
+
+    /// <summary>
+    /// A handler class for configuring custom unlocking conditions for item blueprints.
+    /// </summary>
+    public class KnownTechHandler : IKnownTechHandler
+    {
+        private static readonly KnownTechHandler singleton = new KnownTechHandler();
+
+        public static IKnownTechHandler Main => singleton;
```
AlexejheroYTB: Some handlers only have `Main`, others have both `singleton` and `Main`.

QModManager/API/SMLHelper/Initializer.cs
```diff
@@ -0,0 +1,51 @@
...
+        {
+            Logger.Initialize();
+
+            Logger.Log($"Loading v{Assembly.GetExecutingAssembly().GetName().Version}...", LogLevel.Info);
+
+            harmony = HarmonyInstance.Create("com.ahk1221.smlhelper");
```
AlexejheroYTB: Will this remain the same?

QModManager/API/SMLHelper/Interfaces/ICustomFishHandler.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace QModManager.API.SMLHelper.Interfaces
+{
+    using QModManager.API.SMLHelper.Assets;
+
+    public interface ICustomFishHandler
+    {
+        TechType RegisterFish(CustomFish fish);
```
AlexejheroYTB: xml comment

QModManager/API/SMLHelper/Logger.cs
```diff
@@ -0,0 +1,109 @@
...
+
+    internal enum LogLevel
+    {
+        Debug = 0,
+        Info = 1,
+        Warn = 2,
+        Error = 3,
+    }
+
+    internal static class Logger
```
AlexejheroYTB: Use qmod logger?

QModManager/API/SMLHelper/MonoBehaviours/Fixer.cs
```diff
@@ -0,0 +1,56 @@
...
+        [SerializeField]
+        public TechType techType;
+
+        [SerializeField]
+        public string ClassId;
+
+        private float time;
+        private bool initalized;
+
+        private static FieldInfo BuilderPrefab = typeof(Builder).GetField("prefab", BindingFlags.NonPublic | BindingFlags.Static);
```
AlexejheroYTB: Publicized assembly...?

QModManager/API/SMLHelper/Assets/CustomFishPrefab.cs
PrimeSonic: I am always of the opinion that `using` statements are best kept inside the namespace.  
1. They can be simplified for namespaces inside the same assembly
2. They are harmless
3. Reduced need for aliases
4. For that 0.1% of cases where it will matter, getting into this habit will save you some disambiguation hell

QModManager/API/SMLHelper/Patchers/CraftTreePatcher.cs
```diff
@@ -0,0 +1,231 @@
...
+        internal static void Patch(HarmonyInstance harmony)
+        {
+            harmony.Patch(AccessTools.Method(typeof(CraftTree), "GetTree"),
+                prefix: new HarmonyMethod(AccessTools.Method(typeof(CraftTreePatcher), "GetTreePreFix")));
+
+            harmony.Patch(AccessTools.Method(typeof(CraftTree), "Initialize"),
+                postfix: new HarmonyMethod(AccessTools.Method(typeof(CraftTreePatcher), "InitializePostFix")));
+
+            harmony.Patch(AccessTools.Method(typeof(CraftTree), "FabricatorScheme"),
+                postfix: new HarmonyMethod(AccessTools.Method(typeof(CraftTreePatcher), "FabricatorSchemePostfix")));
```
AlexejheroYTB: Easily accessible using publicized assemblies

QModManager/API/SMLHelper/Patchers/CraftTreePatcher.cs
```diff
@@ -0,0 +1,231 @@
...
+            AddCustomTabs(ref __result, TabNodes, CraftTree.Type.Workbench);
+            PatchNodes(ref __result, CraftingNodes, CraftTree.Type.Workbench);
+        }
+
+        private static void SeamothUpgradesSchemePostfix(ref CraftNode __result)
+        {
+            RemoveNodes(ref __result, NodesToRemove, CraftTree.Type.SeamothUpgrades);
+            AddCustomTabs(ref __result, TabNodes, CraftTree.Type.SeamothUpgrades);
+            PatchNodes(ref __result, CraftingNodes, CraftTree.Type.SeamothUpgrades);
+        }
```
AlexejheroYTB: Can be merged into a single method

QModManager/API/SMLHelper/Patchers/CustomFishPatcher.cs
```diff
@@ -0,0 +1,66 @@
...
+            if(tag)
+            {
+                if(CustomFishHandler.fishTechTypes.Contains(tag.type))
+                {
+                    return;
+                }
+            }
+            if(Random.value < 0.1f)
+            {
+                Logger.Log($"[FishFramework] Selecting fish out of {CustomFishHandler.fishTechTypes.Count} total types", LogLevel.Debug);
```
AlexejheroYTB: Not really necessary

QModManager/API/SMLHelper/Patchers/OptionsPanelPatcher.cs
```diff
@@ -0,0 +1,98 @@
...
+                }
+            }
+            // If no tab was found, create one
+            if (modsTab == -1)
+            {
+                modsTab = optionsPanel.AddTab("Mods");
+            }
+
+            // Maybe this could be split into its own file to handle smlhelper options, or maybe it could be removed alltogether
+            optionsPanel.AddHeading(modsTab, "SMLHelper");
```
AlexejheroYTB: merge with qmodmanager?

QModManager/API/SMLHelper/Patchers/TooltipPatcher.cs
```diff
@@ -0,0 +1,241 @@
...
+            // This is for something else I am going to do
+
+            if (TechTypeHandler.TechTypesAddedBy.TryGetValue(type, out Assembly assembly))
+            {
+                // SMLHelper can access QModManager internals because of this: 
+                // https://github.com/QModManager/QModManager/blob/releases/2.0.1/QModManager/Properties/AssemblyInfo.cs#L21
+
+                string modName = null;
+
+                foreach (QMod mod in Patcher.loadedMods)
```
AlexejheroYTB: Better way of doing this now

QModManager/API/SMLHelper/Utility/ItemStorageHelper.cs
```diff
@@ -0,0 +1,332 @@
...
+
+        #region Common Item Sizes
+
+        private static readonly Vector2int Size1x1 = new Vector2int(1, 1);
+        private static readonly Vector2int Size1x2 = new Vector2int(1, 2);
+        private static readonly Vector2int Size2x1 = new Vector2int(2, 1);
+        private static readonly Vector2int Size2x2 = new Vector2int(2, 2);
+        private static readonly Vector2int Size2x3 = new Vector2int(2, 3);
+        private static readonly Vector2int Size3x1 = new Vector2int(3, 1);
+        private static readonly Vector2int Size3x2 = new Vector2int(3, 2);
```
AlexejheroYTB: What about custom sizes like 4x2?

QModManager/API/SMLHelper/Utility/ReflectionHelper.cs
```diff
@@ -0,0 +1,192 @@
+﻿namespace QModManager.API.SMLHelper.Utility
+{
+    using System.Diagnostics;
+    using System.Reflection;
+
+    /// <summary>
+    /// An extension method class for simplifying reflection calls for improved readability.
+    /// </summary>
+    public static class ReflectionHelper
```
AlexejheroYTB: Has lost its purpose

QModManager/API/SMLHelper/Handlers/KnownTechHandler.cs
PrimeSonic: This was necessary as in some of the handlers it was necessary to access the actual class backing the public interface.  
For those handlers where everything could be done with just the public interface, `Main` alone was kept.  
For handlers where we needed to keep a private instance of the actual class, `singleton` was used.

QModManager/API/SMLHelper/Initializer.cs
PrimeSonic: Why don't we use `"com.qmodmanager.smlhelper"`?

QModManager/API/SMLHelper/Assets/CustomFishPrefab.cs
AlexejheroYTB: When adding the using statement automatically, VS adds it completely and doesn't detect that it could be simplified. This means that we get stuff like this:

```csharp
using QModManager.SMLHelper.Assets; // could be using Assets;
using Interfaces
```

and we have to manually edit it. 

Another argument is that simplified using statements are more likely to break when renaming (CTRL+R, CTRL+R) namespaces.

The only real difference is that they have higher priority over external namespaces, a thing which can be changed in `Code Editing > C# > Formatting Style > Namespace Imports`.

QModManager/API/SMLHelper/Logger.cs
PrimeSonic: The SMLHelper logger and QMod logger should both be fairly similar.  
Yes. We should unify them.

QModManager/API/SMLHelper/MonoBehaviours/Fixer.cs
PrimeSonic: Yes. SMLHelper used the publicized assembly wherever possible.

QModManager/API/SMLHelper/Handlers/KnownTechHandler.cs
AlexejheroYTB: Alright. Some are `singleton` and some are `Singleton` though (I have not been diagnosed with OCD (yet))

QModManager/API/SMLHelper/Initializer.cs
AlexejheroYTB: Why don't we use the same harmony instance?
Also the prefix com is usually used when you have a website, in this case `qmodmanager.com`. The harmony instance that qmodmanager uses is simply `qmodmanager`

QModManager/API/SMLHelper/Patchers/CraftTreePatcher.cs
PrimeSonic: Probably not since these aren't instances of accessing a field or property but rather getting a `MethodInfo` for Harmony patching.

QModManager/API/SMLHelper/MonoBehaviours/Fixer.cs
AlexejheroYTB: This could be easily converted to use publicized assemblies

QModManager/API/SMLHelper/Patchers/OptionsPanelPatcher.cs
PrimeSonic: Yes. Merge items from the Mod Options menu for SMLHelper and QMM together.

QModManager/API/SMLHelper/Assets/CustomFishPrefab.cs
```diff
@@ -0,0 +1,123 @@
...
+        /// <summary>
+        /// The interval in seconds between when your creature finds a new spot to move to
+        /// </summary>
+        public float swimInterval;
+
+        public CustomFishPrefab(string classId, string prefabFileName, TechType techType = TechType.None) : base(classId, prefabFileName, techType)
+        {
+        }
+
+        public override GameObject GetGameObject()
```
AlexejheroYTB: Shouldnt this be GetGameObjectInternal?

QModManager/API/SMLHelper/Utility/ItemStorageHelper.cs
PrimeSonic: Have not encountered that yet. But the cache system here should be able to handle it anyways since the check for "4x2" would still be evaluated.  
These pre-fetched sizes are simply to quickly include known sizes in so we can avoid further calls into the "HasRoomFor" method.

The normal fetching and caching should still happen even for something like a 4x2 item.

QModManager/API/SMLHelper/Utility/ReflectionHelper.cs
PrimeSonic: For us, maybe.  
But this was also offered as a library for other people to use if they wanted.  
If we remove it, we should keep the code somewhere accessible as a reference tool for others.

QModManager/API/SMLHelper/Assets/CustomFishPrefab.cs
PrimeSonic: No. This is correct. `CustomFishPrefab` inherits from `ModPrefab`.  
Nothing outside of `ModPrefab` itself should be calling its `GetGameObjectInternal` method.

AlexejheroYTB: What if an user decides to override this GetGameObject method?

QModManager/API/SMLHelper/Initializer.cs
PrimeSonic: I forgot QMM had its own Harmony instance.  
Sure. Make them the same.

QModManager/API/SMLHelper/Assets/CustomFishPrefab.cs
PrimeSonic: @kylinator25 Consider making this a `sealed override` if you don't intend for other users to make further overrides of this method.  
Otherwise, this is one of those times where any override of the method should include a call into the base implementation.

QModManager/API/SMLHelper/Utility/ReflectionHelper.cs
AlexejheroYTB: It's already in QModManager...

QModManager/API/SMLHelper/Assets/CustomFishPrefab.cs
AlexejheroYTB: This is not very important, but IMO they should be normalized, either per-project, or per-solution.

Examples/Mod.cs
```diff
@@ -0,0 +1,76 @@
+﻿using QModManager.API.SMLHelper.Handlers;
+using QModManager.API.SMLHelper.Options;
+using QModManager.API.SMLHelper.Utility;
+using UnityEngine;
+
+namespace QModManager.Examples
+{
+    public static class ExampleMod
```
AlexejheroYTB: This could also include some QModAPI calls

Examples/Mod.cs
```diff
@@ -0,0 +1,76 @@
...
+            ChoiceIndex = PlayerPrefs.GetInt("SMLHelperExampleModChoice", 0);
+            KeybindKey = PlayerPrefsExtra.GetKeyCode("SMLHelperExampleModKeybind", KeyCode.X);
+            SliderValue = PlayerPrefs.GetFloat("SMLHelperExampleModSlider", 50f);
+            ToggleValue = PlayerPrefsExtra.GetBool("SMLHelperExampleModToggle", true);
+        }
+    }
+
+    public class Options : ModOptions
+    {
+        public Options() : base("SMLHelper Example Mod")
```
AlexejheroYTB: Maybe rename some keys?

QModManager/API/SMLHelper/Initializer.cs
```diff
@@ -0,0 +1,67 @@
+﻿namespace QModManager.API.SMLHelper
```
AlexejheroYTB: We also need to provide an "empty" version of `SMLHelper.dll` with all calls redirected to the new API. I wonder what we will do with V1.

QModManager/API/SMLHelper/Handlers/CraftDataHandler.cs
```diff
@@ -0,0 +1,413 @@
+﻿namespace QModManager.API.SMLHelper.Handlers
+{
+    using Crafting;
+    using Patchers;
+    using Interfaces;
+
+    /// <summary>
+    /// A handler class for adding and editing crafted items.
+    /// </summary>
+    public class CraftDataHandler : ICraftDataHandler
```
AlexejheroYTB: Suggestion: `TechTypeHandler.AddTechType` could return an instance of a `CustomTechType` class, and this class could have all `CraftDataHandler` calls (plus some others) as methods which are called directly on the object.

PrimeSonic: I can see where you're coming from. As always, this will be something added for convenience but won't fully replace the original methods.  

Since people are really embracing the Asset classes for quickly handling the boilerplate for `Craftable` and `Buildable` items, will this new class integrate nicely with those or will it have to be fully separate?   
Let's take a moment to see if rather than create a new class, we can either add these methods onto  the `Spawnable` class itself or via some extension methods so we don't need to add yet another thing to keep track of.

QModManager/API/SMLHelper/Assets/CustomFishPrefab.cs
PrimeSonic: I agree. They should absolutely be normalized, whichever approach we end up with.

QModManager/API/SMLHelper/Initializer.cs
PrimeSonic: I am of the opinion that this should be done as part of the final update to the original SMLHelper project/repo.  
It will be something to release along side the QMM 3.0 update with plenty of `[Obsolete]` tags indicating that this library is for backwards compatibility only and that modders should move to the one in the QMM assembly.

AlexejheroYTB: Agreed, that would have to be done at the end of this pull request after everything else has been finished. About the old repository, a simple readme noting the merge could do it.

QModManager/API/SMLHelper/Assets/PDAItem.cs
```diff
@@ -0,0 +1,67 @@
...
+        private void PatchTechDataEntry()
+        {
+            CraftDataHandler.SetTechData(this.TechType, GetBlueprintRecipe());
+
+            CraftDataHandler.AddToGroup(this.GroupForPDA, this.CategoryForPDA, this.TechType);
+
+            if (this.UnlockedAtStart)
+                KnownTechHandler.UnlockOnStart(this.TechType);
+            else
+                KnownTechHandler.SetAnalysisTechEntry(this.RequiredForUnlock, new TechType[1] { this.TechType }, $"{this.FriendlyName} blueprint discovered!");
```
AlexejheroYTB: This could support custom per-item messages & language overrides.

QModManager/API/SMLHelper/Assets/Spawnable.cs
```diff
@@ -0,0 +1,125 @@
...
+            PrefabHandler.RegisterPrefab(this);
+
+            string assetsFolder = this.AssetsFolder;
+            if (string.IsNullOrEmpty(assetsFolder))
+            {
+                Logger.Error($"AssetsFolder property for Spawnable instance of {this.ClassID} must have a non-empty value.");
+                throw new Exception($"Error patching Spawnable:{this.ClassID}");
+            }
+
+            SpriteHandler.RegisterSprite(this.TechType, $"./QMods/{assetsFolder.Trim('/')}/{this.IconFileName}");
```
AlexejheroYTB: Oh god. Making paths with string literals is a no-no.

QModManager/API/SMLHelper/Crafting/TechData.cs
```diff
@@ -0,0 +1,146 @@
...
+        public TechData(List<Ingredient> ingredients)
+        {
+            Ingredients = ingredients;
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="TechData"/> class for a custom recipe with a collection of ingridients.
+        /// </summary>
+        /// <param name="ingredients">The ingredients.</param>
+        public TechData(params Ingredient[] ingredients)
```
AlexejheroYTB: new constructor idea: `public TechData(params TechType[] types)`

QModManager/API/SMLHelper/Handlers/CraftTreeHandler.cs
```diff
@@ -0,0 +1,339 @@
...
+                case CraftTree.Type.MapRoom:
+                case CraftTree.Type.Centrifuge:
+                case CraftTree.Type.CyclopsFabricator:
+                case CraftTree.Type.Rocket:
+                    break; // Okay
+                case CraftTree.Type.Unused1:                    
+                case CraftTree.Type.Unused2:                    
+                case CraftTree.Type.None:
+                default:
+                    throw new ArgumentException($"{nameof(craftTree)} value of '{craftTree}' does not correspond to a standard crafting tree.{Environment.NewLine}" +
```
AlexejheroYTB: We need to talk about `Enivonment.NewLine`

QModManager/API/SMLHelper/Initializer.cs
```diff
@@ -0,0 +1,48 @@
...
+            ResourcesPatcher.Patch(harmony);
+            PrefabDatabasePatcher.Patch(harmony);
+            SpritePatcher.Patch();
+            KnownTechPatcher.Patch(harmony);
+            BioReactorPatcher.Patch(harmony);
+            OptionsPanelPatcher.Patch(harmony);
+            ItemsContainerPatcher.Patch(harmony);
+            PDAPatcher.Patch(harmony);
+            ItemActionPatcher.Patch(harmony);
+            //TooltipPatcher.Patch(harmony);
```
AlexejheroYTB: TooltipPatcher should be fixed now

QModManager/API/SMLHelper/Initializer.cs
```diff
@@ -0,0 +1,48 @@
...
+        {
+            CustomFishPatcher.Patch(harmony);
+            TechTypePatcher.Patch(harmony);
+            CraftTreeTypePatcher.Patch(harmony);
+            CraftDataPatcher.Patch(harmony);
+            CraftTreePatcher.Patch(harmony);
+            DevConsolePatcher.Patch(harmony);
+            LanguagePatcher.Patch(harmony);
+            ResourcesPatcher.Patch(harmony);
+            PrefabDatabasePatcher.Patch(harmony);
```
AlexejheroYTB: All those patches and all those logs...
This could be improved (i have an idea)

QModManager/API/SMLHelper/Patchers/OptionsPanelPatcher.cs
```diff
@@ -0,0 +1,80 @@
...
+                {
+                    switch (option.Type)
+                    {
+                        case ModOptionType.Slider:
+                            var slider = (ModSliderOption)option;
+
+                            optionsPanel.AddSliderOption(modsTab, slider.Label, slider.Value, slider.MinValue, slider.MaxValue, slider.Value,
+                                new UnityAction<float>((float sliderVal) =>
+                                    modOption.OnSliderChange(slider.Id, sliderVal)));
+                            break;
```
AlexejheroYTB: Bit complicated, maybe split into methods?

QModManager/API/SMLHelper/Patchers/TechTypePatcher.cs
```diff
@@ -0,0 +1,180 @@
...
+                {
+                    __result = techType;
+                    return false;
+                }
+            }
+
+            return true;
+        }
+
+        private static bool Prefix_ToString(Enum __instance, ref string __result)
```
AlexejheroYTB: This could also patch non-custom TechTypes to call .AsString on them

QModManager/API/SMLHelper/Initializer.cs
PrimeSonic: Creating the pass-through assembly will be easiest if we start with the original master of SMLHelper and just rip out all the functionality and replace it with redirects into the QMM assembly.  

The public method calls would be easy. We would need to set up some type redirects for the Asset classes.

QModManager/API/SMLHelper/Utility/EnumCacheManager.cs
```diff
@@ -0,0 +1,210 @@
+﻿namespace QModManager.API.SMLHelper.Utility
+{
+    using QModManager.Utility;
+    using System;
+    using System.Collections.Generic;
+    using System.IO;
+    using System.Reflection;
+    using System.Text;
+
+    internal class EnumTypeCache
```
AlexejheroYTB: Preferably, missing entries should not be deleted.
Missing TechTypes should not be removed from the cache, and SMLHelper should create a dummy item at that index to prevent it from being overridden. Same for CraftTreeTypes

QModManager/API/SMLHelper/Utility/PrefabUtils.cs
```diff
@@ -0,0 +1,22 @@
...
+            var rb = _object.AddComponent<Rigidbody>();
+            _object.AddComponent<PrefabIdentifier>().ClassId = classId;
+            _object.AddComponent<LargeWorldEntity>().cellLevel = LargeWorldEntity.CellLevel.Near;
+            var rend = _object.GetComponentInChildren<Renderer>();
+            rend.material.shader = Shader.Find("MarmosetUBER");
+            var applier = _object.AddComponent<SkyApplier>();
+            applier.renderers = new Renderer[] { rend };
+            applier.anchorSky = Skies.Auto;
+            var forces = _object.AddComponent<WorldForces>();
+            forces.useRigidbody = rb;
```
AlexejheroYTB: Cluttered.

QModManager/API/SMLHelper/Utility/SaveUtils.cs
```diff
@@ -0,0 +1,17 @@
...
+    /* TODO: Maybe add more saving related functions here,
+       such as those related to serializing? */
+    public static class SaveUtils
+    {
+        /// <summary>
+        /// Returns the path to the current save slot's directory.
+        /// </summary>
+        public static string GetCurrentSaveDataDir()
+        {
+            return Path.Combine(SNUtils.savedGamesDir, Utils.GetSavegameDir());
```
AlexejheroYTB: In https://github.com/PrimeSonic/PrimeSonicSubnauticaMods/commit/f748e904186367f26159f2568be918803b60fc6b, you replaced this with something else. Should this also be replaced here?

QModManager/API/SMLHelper/Utility/SelfCheckingDictionary.cs
```diff
@@ -0,0 +1,155 @@
...
+    using System.Collections;
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// This dictionary strtucture automatically checks for duplicate keys as they are being added to the collection.
+    /// Duplicate entires are logged and removed from the final collection.
+    /// </summary>
+    /// <typeparam name="K"></typeparam>
+    /// <typeparam name="V"></typeparam>
+    internal class SelfCheckingDictionary<K, V> : IDictionary<K, V>
```
AlexejheroYTB: Common notation is `<TKey, TValue>`

QModManager/Patcher.cs
```diff
@@ -19,6 +19,8 @@ namespace QModManager
     /// </summary>
     public static class Patcher
     {
+        internal static HarmonyInstance harmony;
+
         internal const string IDRegex = "[^0-9a-z_]";
```
AlexejheroYTB: turn into Regex class

QModManager/Utility/Logger.cs
```diff
@@ -16,13 +17,40 @@ internal static bool EnableDebugging
...
-        internal static void Log(string text) => Log("", text);
-        internal static void Debug(string text) { if (EnableDebugging) Log("/DEBUG", text); }
-        internal static void Info(string text) => Log("/INFO", text);
-        internal static void Warn(string text) => Log("/WARN", text);
-        internal static void Error(string text) => Log("/ERROR", text);
-        internal static void Exception(Exception e) => Log("/EXCEPTION", e.ToString());
-        internal static void Fatal(string text) => Log("/FATAL", text);
+        internal static void Log(string prefix, string text)
+        {
+            Console.WriteLine($"[QModManager{prefix}] {text}");
```
AlexejheroYTB: How should we handle SMLHelper's logging?
I would suggest changing the format to: `[QModManager] [<FROM>] [<LOGLEVEL>]`

QModManager/API/SMLHelper/Assets/PDAItem.cs
PrimeSonic: Absolutely.  
This can be done with a `virtual string` property that defaults to returning `$"{this.FriendlyName} blueprint discovered!");` but can be overridden to use a different message for unlocks.

QModManager/API/SMLHelper/Assets/Spawnable.cs
PrimeSonic: I agree. Let's see if it's possible to have this be backwards compatible for all the mods currently using it which put all these assets into a separate folder that's at the same level as the DLL.

QModManager/API/SMLHelper/Crafting/TechData.cs
PrimeSonic: Assuming you mean `public TechData(params TechType[] ingredients)`, sure.  
That would be fine for many simple recipe scenarios.

QModManager/API/SMLHelper/Handlers/CraftTreeHandler.cs
PrimeSonic: This is for logging, not for printing on screen.  
It makes sense to use Environment.NewLine here since the text will be handled by the OS, not by Unity.

QModManager/API/SMLHelper/Utility/SaveUtils.cs
PrimeSonic: That change is only relevant to Below Zero.  
Any change applied here will need to make the distinction since Subnautica and BZ have different save file locations.

QModManager/API/SMLHelper/Handlers/CraftTreeHandler.cs
AlexejheroYTB: That's why I said we need to talk. The only reason Windows still uses \r\n is for backwards compatibility. Windows doesn't care if we use \n or \r\n, it's the same thing.

QModManager/Patcher.cs
PrimeSonic: Remember that `RegexOptions.Compiled` is not available to .NET 3.5

QModManager/Utility/Logger.cs
PrimeSonic: I am fine with this approach.   
```[QModManager] [<CallingAssemblyName>] [<LOGLEVEL>]```  
I am aware that what were once SMLHelper logs will now be QModManager logs but I am okay with that.

File: QModManager/API/SMLHelper/Initializer.cs (hunk 34 above)
Comment (PrimeSonic) [reply QModManager/API/SMLHelper/Initializer.cs]: It is true that this original structure is a relic of when SMLHelper exposed its patchers directly.  
Since the patchers themselves are no longer exposed publicly, there is an opportunity to combine and consolidate as needed.

Just keep readability in mind so that we don't end up with overly large files that are impossible to find anything in.

File: QModManager/Utility/Logger.cs (hunk 42 above)
Comment (AlexejheroYTB) [reply QModManager/Utility/Logger.cs]: By FROM I dont mean Calling Assembly Name. I mean where the log originated from. (Patcher, VersionCheck, SMLHelper, etc)

File: QModManager/API/SMLHelper/Patchers/OptionsPanelPatcher.cs (hunk 35 above)
Comment (PrimeSonic) [reply QModManager/API/SMLHelper/Patchers/OptionsPanelPatcher.cs]: That _is_ a method call.  
If this is to be made simpler and easier to reach, we should consider combining all those parameters into a single class that can be passed directly.

Comment (AlexejheroYTB) [reply QModManager/API/SMLHelper/Patchers/OptionsPanelPatcher.cs]: No, that is a patch, not a method call. By method calls i mean separate methods for each type of option

File: QModManager/API/SMLHelper/Patchers/TechTypePatcher.cs (hunk 36 above)
Comment (PrimeSonic) [reply QModManager/API/SMLHelper/Patchers/TechTypePatcher.cs]: Hold off on this idea for now. There could be potential here for both useful things and also outright breaking things.  
Let's return to this one after we've sorted out some of the other topics we're trying to address as part of this merge.

File: QModManager/API/SMLHelper/Utility/EnumCacheManager.cs (hunk 37 above)
Comment (PrimeSonic) [reply QModManager/API/SMLHelper/Utility/EnumCacheManager.cs]: I agree that removing missing entries can cause all sorts of problems. Let me propose this:

1. These files are updated with a new format ```{Index}:{Name}:{Active|Inactive}```
2. Whenever a missing custom TechType is detected, we leave the name of the entry alone and instead simply update a bool flag at the end to mark it as inactive.
  a. This flag would be used whenever mods call into the "TryGetModdedTechType" methods,  
to ensure that we don't send back an unusable value.
3. The flag can be switched back to the active state if we see it again on an "AddTechType" call.

This work may have to wait until we've finished merging and be done as another update which may be released along with QMM 3.0 or after it as an update.

File: QModManager/API/SMLHelper/Utility/PrefabUtils.cs (hunk 38 above)
Comment (PrimeSonic) [reply QModManager/API/SMLHelper/Utility/PrefabUtils.cs]: Cluttered, yes. Though I'm not sure how this might be cleaned up beyond just adding some more line breaks for readability.

File: QModManager/API/SMLHelper/Utility/EnumCacheManager.cs (hunk 37 above)
Comment (AlexejheroYTB) [reply QModManager/API/SMLHelper/Utility/EnumCacheManager.cs]: I agree, the cache system would have to be totally re-written for this.

File: QModManager/API/SMLHelper/Handlers/CraftTreeHandler.cs (hunk 32 above)
Comment (PrimeSonic) [reply QModManager/API/SMLHelper/Handlers/CraftTreeHandler.cs]: Sure, Windows can accept both, but we should try to avoid mixing line endings in the same file if we can help it.  
No reason to be _that_ sloppy.

File: QModManager/API/SMLHelper/Handlers/CustomFishHandler.cs (hunk 11 above)
Comment (PrimeSonic) [reply QModManager/API/SMLHelper/Handlers/CustomFishHandler.cs]: Is this specifically about the `CustomFishHandler` class or about all the various `Main` singleton fields?

Comment (AlexejheroYTB) [reply QModManager/API/SMLHelper/Handlers/CustomFishHandler.cs]: Neither. Generally speaking

File: QModManager/API/SMLHelper/Assets/PDAItem.cs (hunk 29 above)
Comment (AlexejheroYTB) [reply QModManager/API/SMLHelper/Assets/PDAItem.cs]: In commit 631e2f1

File: QModManager/API/SMLHelper/Initializer.cs (hunk 34 above)
Comment (AlexejheroYTB) [reply QModManager/API/SMLHelper/Initializer.cs]: I was thinking of attributes, but it would require an int parameter to be passed as the priority, and it would be a bit hard handling those priorities. Unless we can find a way of not needing the priority, this will not be a thing.

File: QModManager/API/SMLHelper/Assets/PDAItem.cs [hunk 43]
```diff
@@ -61,7 +70,7 @@ private void PatchTechDataEntry()
             if (this.UnlockedAtStart)
                 KnownTechHandler.UnlockOnStart(this.TechType);
             else
-                KnownTechHandler.SetAnalysisTechEntry(this.RequiredForUnlock, new TechType[1] { this.TechType }, $"{this.FriendlyName} blueprint discovered!");
+                KnownTechHandler.SetAnalysisTechEntry(this.RequiredForUnlock, new TechType[1] { this.TechType }, Language.main.Get(this.DiscoverMessageKey));
```
PrimeSonic: `Language.main.Get` cannot be used at patch time. It can only be used reliably at run time since we haven't yet patched it with the new usages.  
`SetLanguageLine` only stages lines for patching when the Language `strings` dictionary is initialized (which happens later)

It may be better to simply ensure that the strings sent to `SetAnalysisTechEntry` are being handled correctly for Language Overrides.

QModManager/API/SMLHelper/Assets/PDAItem.cs
```diff
@@ -34,6 +34,13 @@ public abstract class PDAItem : Spawnable
         /// <seealso cref="RequiredForUnlock"/>
         public bool UnlockedAtStart => this.RequiredForUnlock == TechType.None;
 
+        /// <summary>
+        /// Message which should be shown when the item is unlocked
+        /// </summary>
+        public virtual string DiscoverMessage => $"{this.FriendlyName} blueprint discovered!";
```
PrimeSonic: In the XML comments, add a line like we have for the `RequiredForUnlock` property:  
```CSharp
/// <summary>
/// Message which should be shown when the item is unlocked
/// If not overridden, the message will default to $"{FriendlyName} blueprint discovered!".
/// </summary>
```

QModManager/Utility/Logger.cs
AlexejheroYTB: #95

QModManager/API/SMLHelper/Assets/PDAItem.cs
AlexejheroYTB: KnownTech uses the discovery message key as a Language key, so the message returns itself if a key was not found. Changing this to the discover message key should make this work.
(Im a bit tired right now, hopefully what i said makes sense)

QModManager/API/SMLHelper/Assets/PDAItem.cs
```diff
@@ -0,0 +1,77 @@
...
+        private void PatchTechDataEntry()
+        {
+            CraftDataHandler.SetTechData(this.TechType, GetBlueprintRecipe());
+
+            CraftDataHandler.AddToGroup(this.GroupForPDA, this.CategoryForPDA, this.TechType);
+
+            if (this.UnlockedAtStart)
+                KnownTechHandler.UnlockOnStart(this.TechType);
+            else
+                KnownTechHandler.SetAnalysisTechEntry(this.RequiredForUnlock, new TechType[1] { this.TechType }, this.DiscoverMessageKey);
```
PrimeSonic: I see you've changed this method call to pass in the Language strings key instead of the message.  
Did you first confirm that this would work?

Seeing as the commit that made the changes (  0913cb9f34f19f18b80fdfd44730f2964b10ac32) appears as _Verified_, I can't be sure.
Can you confirm that the `unlockMessage` field from `KnownTech.AnalysisTech` is pass through `Language.main.Get` instead of being used directly?

AlexejheroYTB: 0913cb9 is not verified, plus it includes built assemblies.
I am currently unable and unwilling to test since I have not finished the SMLHelper integration, but according to the game's code and KnownTechHandler, it will work.
Some of the methods in KnownTechHandler have a default parameter NotificationBlueprintUnlocked, which is a language key.

PrimeSonic: > Some of the methods in KnownTechHandler have a default parameter NotificationBlueprintUnlocked, which is a language key. 

Works for me 👍

QModManager/API/SMLHelper/Patchers/BioReactorPatcher.cs
```diff
@@ -0,0 +1,20 @@
...
+    using QModManager.Utility;
+    using System.Collections.Generic;
+
+    internal static class BioReactorPatcher
+    {
+        internal static IDictionary<TechType, float> CustomBioreactorCharges = new SelfCheckingDictionary<TechType, float>("CustomBioreactorCharges", TechTypeExtensions.sTechTypeComparer);
+
+        internal static void Patch(HarmonyInstance harmony)
+        {
+            // Direct access to private fields made possible by https://github.com/CabbageCrow/AssemblyPublicizer/
```
AlexejheroYTB: Are these comments necessary?

QModManager/API/SMLHelper/Patchers/CustomFishPatcher.cs
```diff
@@ -0,0 +1,65 @@
...
+            Type creatureType = typeof(Creature);
+            Type thisType = typeof(CustomFishPatcher);
+
+            harmony.Patch(creatureType.GetMethod("Start", BindingFlags.Public | BindingFlags.Instance),
+                null, new HarmonyMethod(thisType.GetMethod("CreatureStart_Postfix", BindingFlags.NonPublic | BindingFlags.Static)), null);
+
+            Logger.Debug("CustomFishPatcher is done.");
+        }
+
+        private static void CreatureStart_Postfix(Creature __instance)
```
AlexejheroYTB: (note) A bit of cleanup here would be nice

QModManager/API/SMLHelper/Patchers/DevConsolePatcher.cs
```diff
@@ -0,0 +1,60 @@
...
+
+            if (args.Length != 0)
+            {
+                foreach (CommandInfo command in commands)
+                {
+                    if (command.Name.Contains(args[0]))
+                    {
+                        List<string> argsList = args.ToList();
+                        argsList.RemoveAt(0);
+                        string[] newArgs = argsList.ToArray();
```
AlexejheroYTB: (note) This is a bit more complicated than it should be.

QModManager/API/SMLHelper/Utility/ExtBannedIdManager.cs
```diff
@@ -0,0 +1,141 @@
...
+    /// <summary>
+    /// This class is tasked with checking external txt files for banned IDs that are not to be isssued when patching in new enum entries.
+    /// </summary>
+    internal static class ExtBannedIdManager
+    {
+        private static bool IsInitialized = false;
+
+        private static readonly Dictionary<string, List<int>> BannedIdDictionary = new Dictionary<string, List<int>>();
+
+        private static readonly string BannedIdDirectory = Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location), "RestrictedIDs");
```
AlexejheroYTB: All path calls will need to be changed after the merge. I will handle this in #81, since it will also have path changes.

QModManager/Patcher.cs
```diff
@@ -83,14 +83,13 @@ internal static void Patch()
 
                 VersionCheck.Check();
 
-                QModHooks.Start += PrefabDebugger.Main;
+                //QModHooks.Start += PrefabDebugger.Main;
+
+                Initializer.PostPostInit();
```
AlexejheroYTB: (note) We need to not load SMLHelper as a regular mod

QModManager/Utility/IOUtilities.cs
```diff
@@ -31,7 +31,7 @@ internal static string GetFolderStructureAsTree(string directory = null)
             }
         }
 
-        internal static string GenerateFolderStructure(string directory)
+        private static string GenerateFolderStructure(string directory)
```
AlexejheroYTB: (note) A lot of `internal` things can be made `private` as long as they're not used in Unit Tests

Unit Tests/SMLHelper/SelfCheckingDictionary.cs
```diff
@@ -4,7 +4,7 @@
 namespace QModManager.Tests.SMLHelper
 {
     [TestFixture]
-    internal class SelfCheckingDictionaryTests
+    internal class SelfCheckingDictionary
```
PrimeSonic: This is a `TestFixture` with unit tests that validate the functionality of the `SelfCheckingDictionary` class.  
"`SelfCheckingDictionaryTests`" is an appropriate name by convention.

Renaming this test class to the same name as the class its testing just looks confusing.

Unit Tests/SMLHelper/LanguagePatcher.cs
```diff
@@ -6,7 +6,7 @@
 namespace QModManager.Tests.SMLHelper
 {
     [TestFixture]
-    public class LanguagePatcherTests
+    public class LanguagePatcher
```
PrimeSonic: Any reason why this needed to be renamed?  This is a collection of unit tests.  
Do you have objections against the naming convention "<class being tested>Tests" for these TestFixtures?

QModManager/API/SMLHelper/Patchers/PDAPatcher.cs
```diff
@@ -24,9 +24,7 @@ internal static void Patch(HarmonyInstance harmony)
         private static void InitializePostfix()
         {
             BlueprintToFragment.Clear();
-
-            // Direct access to private fields made possible by https://github.com/CabbageCrow/AssemblyPublicizer/
```
PrimeSonic: We should either keep these comments or have a link to the AssemblyPublicizer project somewhere visible in the repo.  
Anyone looking to contribute to the SMLHelper APIs of QModManager may very well get confused when they see us directly accessing various private elements without Reflection.

Examples/mod.json
```diff
@@ -0,0 +1,9 @@
+{
+  "Id": "ExampleMod",
+  "DisplayName": "Example Mod",
+  "Author": "the QModManager dev team",
+  "Version": "1.0",
+  "Enable": true,
+  "AssemblyName": "Examples.dll",
+  "EntryMethod": "QModManager.Examples.ExampleMod.Patch"
```
PrimeSonic: While mostly irrelevant, we should consider updating this to the newer format since some modders may want to compile and run it.  
If it's to be used as a guideline, we should keep it up to date.

QModManager/API/SMLHelper/Assets/Buildable.cs
```diff
@@ -0,0 +1,39 @@
...
+            CorePatchEvents += PatchBuildable;
+        }
+
+        private void PatchBuildable()
+        {
+            CraftDataHandler.AddBuildable(this.TechType);
+
+            if (!string.IsNullOrEmpty(HandOverText))
+            {
+                LanguageHandler.SetLanguageLine($"{this.ClassID}HandOver", this.HandOverText);
```
PrimeSonic: I know for a fact that this doesn't work the way I had originally intended.  
For right now, I think the best idea is to purge the entire `HandOverText` property along with this LanguageLine patching.

Unit Tests/SMLHelper/LanguagePatcher.cs
AlexejheroYTB: Nope, just wanted to unify this with QMM's tests which do not include Tests at the end. I guess I'll do the other way around.

QModManager/API/SMLHelper/Assets/ModPrefab.cs
```diff
@@ -0,0 +1,123 @@
...
+
+                if (go.GetComponent<Constructable>() != null)
+                {
+                    go.GetComponent<Constructable>().techType = TechType;
+                }
+            }
+
+            if (go.GetComponent<PrefabIdentifier>() != null)
+            {
+                go.GetComponent<PrefabIdentifier>().ClassId = ClassID;
```
PrimeSonic: I have come across instances where a modded prefab I created did not have the `PrefabIdentifier` component when cloned, making it so that the component had to be added manually in the `override GetGameObject` method.  
Without this component, there isn't really a good way to identify the item when save data is being loaded.

Open question:  
**Should we go ahead and ensure that this component is always added to all `ModPrefab` created gameobjects?**

QModManager/API/SMLHelper/Assets/PDAItem.cs
```diff
@@ -0,0 +1,78 @@
...
+    /// A <see cref="Spawnable"/> item that appears in the PDA blueprints.
+    /// </summary>
+    /// <seealso cref="Spawnable" />
+    public abstract class PDAItem : Spawnable
+    {
+        /// <summary>
+        /// Override to set the <see cref="TechType"/> that must first be scanned or picked up to unlock the blueprint for this item. <para/>
+        /// If not overriden, it this item will be unlocked from the start of the game.
+        /// </summary>
+        public virtual TechType RequiredForUnlock => TechType.None;
```
PrimeSonic: As a later revision, I plan on adding a new property "`virtual IEnumerable<TechType> UnlockedBy => null;`"  
This will accept any list of TechTypes such that when _any_ of those TechTypes are scanned, the `PDAItem` here is unlocked.

This will likely go hand in hand with normalizing the unlocking method that @kylinator25 came up with for attaching blueprint unlocks to Encyclopedia entries rather than other blueprints.

QModManager/API/SMLHelper/Assets/Spawnable.cs
```diff
@@ -0,0 +1,128 @@
...
+        public void Patch()
+        {
+            if (this.IsPatched)
+                return; // Already patched. Skip.
+
+            OnStartedPatching?.Invoke();
+
+            // Because invocation order isn't guaranteed by event handlers,
+            // we make sure the TechType is patched first before anything else that might require it.
+            this.TechType = TechTypeHandler.Singleton.AddTechType(ModName, this.ClassID, this.FriendlyName, this.Description, false);
```
PrimeSonic: This might be something I end up handling.  
Instead of this line here directly adding the TechType, instead it should be put into an `internal virtual` method `PatchTechType` that is overridden by `PDAItem`.  
Within the scope of `PDAItem`, we can pass in the correct "UnlocksAtStart" value.  
This current implementation is fine within `Spawnable` since use of this class alone isn't intended for items that are unlocked as blueprints.

QModManager/API/SMLHelper/Crafting/ModCraftTreeFamily.cs
```diff
@@ -0,0 +1,446 @@
...
+    using Patchers;
+    using QModManager.Utility;
+    using UnityEngine;
+    using UnityEngine.Assertions;
+    using Utility;
+
+    /// <summary>
+    /// Basic data structure of a crafting tree node.
+    /// </summary>
+    public abstract class ModCraftTreeNode
```
PrimeSonic: @AlexejheroYTB If you agree, we can create a new folder `QModManager/API/SMLHelper/Crafting/CraftTreeModding` and split all these classes into their own separate files.

QModManager/API/SMLHelper/Patchers/PDAPatcher.cs
AlexejheroYTB: Documentation will be the last thing for 3.0

QModManager/API/SMLHelper/Handlers/SpriteHandler.cs
```diff
@@ -0,0 +1,221 @@
...
+        /// </summary>
+        /// <param name="type">The techtype paired to this sprite.</param>
+        /// <param name="filePathToImage">The file path to image to be converted into a sprite.</param>
+        /// <param name="format"><para>The texture format. By default, this uses <see cref="TextureFormat.BC7" />.</para>
+        /// <para>https://docs.unity3d.com/ScriptReference/TextureFormat.BC7.html</para>
+        /// <para>Don't change this unless you really know what you're doing.</para></param>
+        /// <seealso cref="ImageUtils.LoadSpriteFromFile(string, TextureFormat)" />
+        void ISpriteHandler.RegisterSprite(TechType type, string filePathToImage, TextureFormat format)
+        {
+            Atlas.Sprite sprite = ImageUtils.LoadSpriteFromFile(filePathToImage, format);
```
PrimeSonic: Heads up: This is breaking in Unity2018.

QModManager/API/SMLHelper/MonoBehaviours/Fixer.cs
```diff
@@ -0,0 +1,55 @@
...
+        [SerializeField]
+        public TechType techType;
+
+        [SerializeField]
+        public string ClassId;
+
+        private float time;
+        private bool initalized;
+
+        private static FieldInfo BuilderPrefab = AccessTools.Field(typeof(Builder), "prefab");
```
PrimeSonic: Can this be accessed directly with the publicized assembly?

QModManager/API/SMLHelper/Patchers/CraftTreePatcher.cs
```diff
@@ -0,0 +1,224 @@
...
+        {
+            SchemePostfix(ref __result, CraftTree.Type.MapRoom);
+        }
+
+        private static void CyclopsFabricatorSchemePostfix(ref CraftNode __result)
+        {
+            SchemePostfix(ref __result, CraftTree.Type.CyclopsFabricator);
+        }
+
+        private static void SchemePostfix(ref CraftNode __result, CraftTree.Type type)
```
PrimeSonic: For clarity, let's rename this new method `PatchCraftingTree`.

QModManager/API/SMLHelper/MonoBehaviours/Fixer.cs
AlexejheroYTB: It can, but we need its `FieldInfo`. If the `FieldInfo` is only used for `.GetValue`, we could replace this line.

QModManager/API/SMLHelper/Handlers/SpriteHandler.cs
AlexejheroYTB: Will be handled later.

Examples/mod.json
AlexejheroYTB: Everything new added to the `mod.json` file is optional.

QModManager/API/SMLHelper/Assets/Buildable.cs
AlexejheroYTB: I'll let you handle this one.

QModManager/API/SMLHelper/Assets/ModPrefab.cs
AlexejheroYTB: I haven't worked a lot with prefabs, so I'll let you handle this one

QModManager/API/SMLHelper/Crafting/ModCraftTreeFamily.cs
AlexejheroYTB: Yea, the file is very cluttered and contains a lot of classes. I'll let you split them up as you want. You can even change the namespaces as we don't need this to be backwards-compatible.

QModManager/API/SMLHelper/Utility/StorageHelperExtensions.cs
```diff
@@ -0,0 +1,96 @@
+﻿namespace QModManager.API.SMLHelper.Utility
+{
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// Extension methods to provide static calls for <see cref="ItemsContainer"/> into <see cref="ItemStorageHelper"/> methods.
+    /// </summary>
+    public static class StorageHelperExtensions
```
AlexejheroYTB: This storage helper's all over the place, it's in Interfaces, it's in Utility, and it's in Utility again. Why so many files?

QModManager/API/SMLHelper/MonoBehaviours/Fixer.cs
```diff
@@ -0,0 +1,55 @@
+namespace QModManager.API.SMLHelper.MonoBehaviours
+{
+    using UnityEngine;
+    using System.Reflection;
+    using Logger = QModManager.Utility.Logger;
+    using Harmony;
+
+    public class Fixer : MonoBehaviour, IProtoEventListener
```
AlexejheroYTB: Is this meant for public usage?

QModManager/API/SMLHelper/Utility/ItemStorageHelper.cs
```diff
@@ -0,0 +1,328 @@
+﻿namespace QModManager.API.SMLHelper.Utility
+{
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// A utility class that offers additional info about <see cref="ItemsContainer"/> instances.
+    /// </summary>
+    /// <seealso cref="IStorageHelper" />
+    public class ItemStorageHelper : IStorageHelper
```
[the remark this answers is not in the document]
PrimeSonic: Mainly because it's exposed for public use.

QModManager/API/SMLHelper/Crafting/ModCraftTreeFamily.cs
```diff
@@ -0,0 +1,446 @@
...
+
+        /// <summary>
+        /// The the in-game representation of this node.
+        /// </summary>
+        public CraftNode CraftNode;
+
+        protected ModCraftTreeLinkingNode Parent = null;
+
+        protected virtual CraftTree.Type Scheme => this.Parent.Scheme;
+        protected virtual string SchemeAsString => this.Parent.SchemeAsString;
```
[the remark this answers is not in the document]
PrimeSonic: These could be `internal` instead.

QModManager/API/SMLHelper/Handlers/BioReactorHandler.cs
```diff
@@ -0,0 +1,34 @@
+﻿namespace QModManager.API.SMLHelper.Handlers
+{
+    using Patchers;
+    using Interfaces;
+
+    public class BioReactorHandler : IBioReactorHandler
+    {
+        public static IBioReactorHandler Main { get; } = new BioReactorHandler();
```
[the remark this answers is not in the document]
PrimeSonic:
```CSharp
/// <summary>
/// Main entry point for all calls to this handler.
/// </summary>
```
Copy & Paste as needed.

QModManager/API/SMLHelper/Utility/StorageHelperExtensions.cs
PrimeSonic: @DingoDjango Did a lot of these so I'm not all that sure.  
The Extensions look to mostly be there for convenience.

QModManager/API/SMLHelper/MonoBehaviours/Fixer.cs
PrimeSonic: I don't believe so, given that we manage adding this component ourselves.  
It could be `internal`, since I don't think we've ever given any guidance to modders to use it directly.

QModManager/API/SMLHelper/Handlers/CraftDataHandler.cs
AlexejheroYTB: These classes already have properties defining these values. I don't think additional methods would be required

QModManager/API/SMLHelper/Assets/PDAItem.cs
```diff
@@ -0,0 +1,78 @@
...
+        /// </summary>
+        /// <value>
+        ///   Returns <c>true</c> if will be unlocked from the start of the game; otherwise, <c>false</c>.
+        /// </value>
+        /// <seealso cref="RequiredForUnlock"/>
+        public bool UnlockedAtStart => this.RequiredForUnlock == TechType.None;
+
+        /// <summary>
+        /// Message which should be shown when the item is unlocked. <para/>
+        /// If not overridden, the message will default to "<see langword="{FriendlyName} blueprint discovered!"/>"
```
AlexejheroYTB: (note) XML comments needs to be updated here

QModManager/API/SMLHelper/Utility/StorageHelperExtensions.cs
DingoDjango: I put everything in the same folder as far as I remember. We had an interface as PS requested, a helper class which did the actual work and a static class which exposed methods to intellisense.

QModManager/API/SMLHelper/Crafting/ModCraftTreeFamily.cs
PrimeSonic: Refactor cannot be done without first moving the SMLHelper shared project outside of the QModManager project folder.

PR #94, opened 2019-05-25
Update 3.0

AlexejheroYTB:
### Changelog

- The project is now licensed under [CC BY-NC-ND 4.0](https://creativecommons.org/licenses/by-nc-nd/4.0/) (c9e7413, #97)
- QModManager has merged with SMLHelper! (#93)
- QModManager has been updated to Unity 2018 to work with the latest version of Subnautica! Older versions of Subnautica may not work! (#99)

---

### To do

- #98 - Feature updates

<wbr>

- #81 - Doorstop injection
- #103 - Use an API for making mods instead of using a `mod.json` file


QModManager/API/SMLHelper/Crafting/ModCraftTreeNode.cs
```diff
@@ -0,0 +1,78 @@
...
+            CraftNode = new CraftNode(Name, Action, TechType);
+        }
+
+        /// <summary>
+        /// Removes the calling node from parent. 
+        /// </summary>
+        public void RemoveNode()
+        {
+            Assert.IsNotNull(this.Parent, "No parent found to remove node from!");
+            Assert.IsNotNull(this.Parent.CraftNode, "No CraftNode found on parent!");
```
AlexejheroYTB: Are asserts required here?

QModManager/API/SMLHelper/Crafting/ModCraftTreeRoot.cs
```diff
@@ -0,0 +1,110 @@
...
+                    CreateFromExistingTree(node, ref thing);
+                }
+
+                if (node.action == TreeAction.Craft)
+                {
+                    TechType techType = TechType.None;
+                    TechTypeExtensions.FromString(node.id, out techType, false);
+
+                    if (node.id == "SeamothHullModule2") techType = TechType.VehicleHullModule2;
+                    else if (node.id == "SeamothHullModule3") techType = TechType.VehicleHullModule3;
```
AlexejheroYTB: Why is this hard-coding here?

QModManager/API/SMLHelper/Handlers/BioReactorHandler.cs
```diff
@@ -0,0 +1,40 @@
+﻿namespace QModManager.API.SMLHelper.Handlers
+{
+    using Patchers;
+    using Interfaces;
+
+    /// <summary>
+    /// A handler class for managing Bio Reactor fuels
+    /// </summary>
+    public class BioReactorHandler : IBioReactorHandler
```
AlexejheroYTB: Custom build events for automatically converting these classes to support dependency injection?

QModManager/API/SMLHelper/Handlers/CraftTreeHandler.cs
```diff
@@ -0,0 +1,338 @@
...
+                case CraftTree.Type.MapRoom:
+                case CraftTree.Type.Centrifuge:
+                case CraftTree.Type.CyclopsFabricator:
+                case CraftTree.Type.Rocket:
+                    break; // Okay
+                case CraftTree.Type.Unused1:                    
+                case CraftTree.Type.Unused2:                    
+                case CraftTree.Type.None:
+                default:
+                    throw new ArgumentException($"{nameof(craftTree)} value of '{craftTree}' does not correspond to a standard crafting tree. This method is intended for use only with standard crafting trees, not custom ones or unused ones.");
```
AlexejheroYTB: Can we do this with `<`/`>`?

QModManager/API/SMLHelper/Patchers/BioReactorPatcher.cs
```diff
@@ -0,0 +1,18 @@
...
+    using QModManager.Utility;
+    using System.Collections.Generic;
+
+    internal static class BioReactorPatcher
+    {
+        internal static IDictionary<TechType, float> CustomBioreactorCharges = new SelfCheckingDictionary<TechType, float>("CustomBioreactorCharges", TechTypeExtensions.sTechTypeComparer);
+
+        internal static void Patch(HarmonyInstance harmony)
+        {
+            PatchUtils.PatchDictionary(BaseBioReactor.charge, CustomBioreactorCharges);
```
AlexejheroYTB: Could we do something to allow patchers to also patch stuff at **run-time**?

QModManager/API/SMLHelper/Utility/ExtBannedIdManager.cs
```diff
@@ -0,0 +1,141 @@
...
+    using System;
+    using System.Collections;
+    using System.Collections.Generic;
+    using System.IO;
+    using System.Reflection;
+
+    /// <summary>
+    /// This class is tasked with checking external txt files for banned IDs that are not to be isssued when patching in new enum entries.
+    /// </summary>
+    internal static class ExtBannedIdManager
```
AlexejheroYTB: Missing entries in the cache should not be removed!

QModManager/API/SMLHelper/Patchers/BioReactorPatcher.cs
PrimeSonic: That would only work for patching of data collections like this.  
For Harmony and TechType patching it wouldn't.

Making the difference apparent to modders using the APIs would be a bit of a challenge, so for the sake of simplicity, it would be easier to just keep them both consistent.

QModManager/API/SMLHelper/Handlers/CraftTreeHandler.cs
PrimeSonic: The invalid cases are `int` values 0, 4, and 5.
```CSharp
public enum Type
{
    None = 0,
    Fabricator = 1,
    Constructor = 2,
    Workbench = 3,
    Unused1 = 4,
    Unused2 = 5,
    SeamothUpgrades = 6,
    MapRoom = 7,
    Centrifuge = 8,
    CyclopsFabricator = 9,
    Rocket = 10
}
```
Normally, the compiler would crush a switch statement like this into a simple math check.  
But in this case, the values to check for aren't in a continuous range.  
So there's no easy math that can be done.

The compiler has done the best that can be done already which is this:
image

QModManager/API/SMLHelper/Crafting/ModCraftTreeRoot.cs
PrimeSonic: Looks like this is here to map the old, unused TechTypes for the Seamoth depth modules with the ones now in use.  
The Cyclops also has some old, unused Depth Upgrade TechTypes.

We could probably remove this entirely.

QModManager/API/SMLHelper/Crafting/ModCraftTreeNode.cs
PrimeSonic: I had added this originally to assist modders, so they could see these errors early.  
Running into these errors would require doing something pretty careless anyways.

QModManager/API/SMLHelper/Assets/ModSprite.cs
```diff
@@ -0,0 +1,112 @@
...
+        /// <summary>
+        /// Creates a new ModSprite to be used with a specific TechType.
+        /// Created with an Atlas Sprite.
+        /// </summary>
+        /// <param name="type">The techtype paired to this sprite.</param>
+        /// <param name="sprite">The sprite to be added.</param>
+        public ModSprite(TechType type, Atlas.Sprite sprite)
+        {
+            TechType = type;
+            Id = type.AsString();
```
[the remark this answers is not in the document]
PrimeSonic: That might result in getting into an infinite loop and eventual StackOverflow.  
image
I wouldn't recommend it.

PR #95, opened 2019-05-28
Update logger

AlexejheroYTB:
Changes won't be very visible if done directly in #93


AlexejheroYTB:
If you approve this merge it directly

PR #101, opened 2019-05-31
Get rid of mod.json and create mods by inheriting from a QMod class/interface

AlexejheroYTB:
This is a very rough concept and I don't think it even builds yet. I will start working on this again after #81 is merged.

This is not very high priority, so I'm putting it down as an "Optional" feature.


AlexejheroYTB:
_(meant to make it a draft, oops)_

AlexejheroYTB:
Closed in favor of #103

PR #103, opened 2019-06-01
Mod loading API

PrimeSonic:
### Mod Loading Overhaul
- Overhauls the `Patcher` class, delegating specific tasks into separate classes
  - The `QModFactory` handles reading json files and assemblies to create the `QMod` instance
    - Makes use of the newly introduced `SortedTree` data structure to check for missing and circular dependencies and load orders
  - The `Initializer` class handles running the patch methods for the QMods under the new patching order
    - Receives reports of the status of patching, in case of errors or unsupported mods
- New Pre/Post Patching options are available for library/API type mods.


QModManager/API/ModLoading/Internal/QMod.cs
```diff
@@ -0,0 +1,70 @@
+﻿namespace QModManager.API.ModLoading.Internal
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Reflection;
+
+    internal class QMod
```
AlexejheroYTB: Rename this to something else so we it doesn't get confused with the other `QMod` class

QModManager/API/ModLoading/QModDependency.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="QModDependency"/> class.
+        /// </summary>
+        /// <param name="requiredMod">The other mod required by this one to function.</param>
+        /// <param name="requiredVersion">The minimum version the other mod should be at.</param>
+        public QModDependency(string requiredMod, string requiredVersion)
+        {
+            this.RequiredMod = requiredMod;
+            this.MinimumVersion = requiredVersion == null ? new Version() : new Version(requiredVersion);
```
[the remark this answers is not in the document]
PrimeSonic: Not possible. See documentation:  
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/attributes#attribute-parameter-types

Best we can do is replace this with a series of `int` parameters instead.

QModManager/API/ModLoading/QModCoreInfo.cs
```diff
@@ -0,0 +1,58 @@
...
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="QModCoreInfo" /> class.
+        /// </summary>
+        /// <param name="id">The mod identifier.</param>
+        /// <param name="displayName">The mod display name.</param>
+        /// <param name="author">The mod author or team name.</param>
+        /// <param name="developedFor">The game the mod was developed for.</param>
+        /// <param name="patchMethod">The patch method.</param>
+        public QModCoreInfo(string id, string displayName, string author, DevelopedFor developedFor, string patchMethod)
```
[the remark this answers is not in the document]
PrimeSonic: The version info can (and IMO should) be pulled from the `AssemblyInfo`.  
I'll be adding that shortly.

QModManager/API/ModLoading/DevelopedFor.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// <summary>
+    /// Defines what game a QMod can run on.
+    /// </summary>
+    [Flags]
+    public enum DevelopedFor
+    {
+        /// <summary>
+        /// Mod developed exclusively for Subnautica
+        /// </summary>
+        Subnautica = Patcher.Game.Subnautica,
```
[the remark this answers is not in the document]
PrimeSonic: The documentation will need to be changed.  
`Game.Both` is indicative of a more flexible mod when provided from this scope.  
But the same value indicates an error condition in the installer space (according to the existing XML summary).

QModManager/API/ModLoading/Internal/QMod.cs
PrimeSonic: My intention is that all these classes can combine into one, or at least share a common interface.  
That I have yet to dig into.

PrimeSonic: Still WiP

QModManager/API/ModLoading/QModPatchMethod.cs
```diff
@@ -0,0 +1,37 @@
...
+    using System;
+    using Advanced;
+
+    /// <summary>
+    /// Identifies a patch method for a QMod.<para/>
+    /// ALERT: The class that defines this method must have a <seealso cref="QModCoreInfo"/> attribute.
+    /// </summary>
+    /// <seealso cref="System.Attribute" />
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = false, Inherited = false)]
+    public class QModPatchMethod : Attribute
```
AlexejheroYTB: Maybe it's time for that `PreInit`/`Init`/`PostInit` type of initialization

PrimeSonic: Already working on it.  
See `QModManager/API/ModLoading/Advanced/MetaPatchOrder.cs`

AlexejheroYTB: That is a replacement for `Priority`, and not what I mean

PrimeSonic: PreInit = Patch Early  
Init = Patch Normal  
PostInit = Patch Late

QModManager/API/ModLoading/Internal/QMod.cs
PrimeSonic: We are back to a single internal `QMod` class.  
There is a `IQMod` interface for public use.  
Other interfaces implemented by this class are for internal use only.

QModManager/API/ModLoading/QModPatchMethod.cs
PrimeSonic: Resolving for now as this is still WiP until the `Patcher` class is redone.  
There are 3 separate attributes: 
- `[QModPatchMethod]`, for 99.9% of mods
- `[QModPostPatchMethod]`, for mods that need something done _after_ all normal patching has completed
- `[QModPrePatchMethod]`, for mods that need something done _before_ all normal patching starts

AlexejheroYTB: But mods can have both 3 at the same time i assume

PrimeSonic: Yes. A mod could have one of each. And I will be adding some additional error checking for it.

For example:  
- Any mod with optional debug logging could use `[QModPrePatchMethod]` to set up logging preferences, then normal patching happens on `[QModPatchMethod]`.
- CustomCraft2 could set up logging preferences in either `[QModPatchMethod]` or `[QModPrePatchMethod]`, because it would do all its real patching in `[QModPostPatchMethod]` to avoid having to keep adding a `[LoadBefore]` attribute for every other mod.

Note that anything executing in `[QModPostPatchMethod]` would still run _before_ the SMLHelper patching.

QModManager/Patcher.cs
```diff
@@ -37,10 +37,10 @@ internal static string QModBaseDir
         }
         private static bool Patched = false;
```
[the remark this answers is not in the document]
PrimeSonic: Overhaul complete. Solution can build again. Testing pending.

QModManager/DataStructures/SortedTree.cs
```diff
@@ -0,0 +1,217 @@
...
+                    NodesInError++;
+                    break;
+            }
+
+            return sortResult;
+        }
+
+        public List<IdType> GetErrors()
+        {
+            return erroredIds;
```
PrimeSonic: Maybe this should be an `out` parameter on the `CreateFlatIndexList` and `CreateFlatValueList` methods.

QModManager/DataStructures/SortedTreeNode.cs
```diff
@@ -0,0 +1,313 @@
...
+                        if (testAfterResult > SortResults.NoSortPreference)
+                        {
+                            return testAfterResult;
+                        }
+
+                        midLevelResult = testAfterResult;
+                    }
+                    else if (this.LoadAfter == null && this.LoadBefore != null)
+                    {
+                        SortResults testBeforeResult = SortBefore(other, true);
```
PrimeSonic: Replace literal bool values with named constants for clarity.

QModManager/API/ModLoading/IQMod.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        /// <summary>
+        /// A list of mods, before which, this mod will load
+        /// </summary>
+        string[] LoadBefore { get; }
+
+        /// <summary>
+        /// A list of mods, after which, this mod will load
+        /// </summary>
+        string[] LoadAfter { get; }
```
PrimeSonic: Ideally, this would be some `readonly` type of collection.  
This possible now since `IQModSerializable` enforces the requirements for the old json files.

Executable/AudioFixer.cs
```diff
@@ -1,20 +1,21 @@
...
+    using System.Collections.Generic;
+    using System.IO;
+
     internal static class AudioFixer
     {
-        internal static void ChangeDisableUnityAudio(string path, bool newValue, Patcher.Game game)
+        internal static void ChangeDisableUnityAudio(string path, bool newValue, Game game)
         {
-            if (game != Patcher.Game.Subnautica && game != Patcher.Game.BelowZero)
+            if (game != Game.Subnautica && game != Game.BelowZero)
```
PrimeSonic: Could be simplified as `game == Game.None`

QModManager/API/Game.cs
```diff
@@ -0,0 +1,31 @@
+﻿namespace QModManager.API
+{
+    using System;
+
+    /// <summary>
+    /// Identifies the Subnautica games.
+    /// </summary>
+    [Flags]
+    public enum Game
```
PrimeSonic: Could we use a less ambiguous name for this enum?

QModManager/API/ModLoading/Internal/PatchMethod.cs
```diff
@@ -0,0 +1,52 @@
...
+        internal string ModId { get; }
+        internal MethodInfo Method { get; }
+        internal Assembly LoadedAssembly { get; }
+        internal bool IsPatched { get; private set; }
+
+        internal bool TryInvoke()
+        {
+            try
+            {
+                this.Method.Invoke(this.LoadedAssembly, new object[] { });
```
PrimeSonic: This currently only works for `static` methods.  
With a little extra work, this and the supporting classes could be expanded to support non-static methods as well.

@AlexejheroYTB Do you see any benefit allowing this?  
The main benefit I see is the added flexibility would make it harder for first-time modders to "get it wrong".

QModManager/API/ModLoading/Internal/Initializer.cs
```diff
@@ -0,0 +1,178 @@
...
+            where Q : IQModLoadable
+        {
+            InitializeMods(modsToInitialize, PatchingOrder.PreInitialize);
+            InitializeMods(modsToInitialize, PatchingOrder.NormalInitialize);
+            InitializeMods(modsToInitialize, PatchingOrder.PostInitialize);
+            FinalInitialize();
+
+            LogResults(ModLoadingResults.Failure, "The following mods failed to load to the errors during their initialization");
+            LogResults(ModLoadingResults.AlreadyLoaded, "The following mods encountered duplicate initialization attempts");
+            LogResults(ModLoadingResults.CurrentGameNotSupported, $"The following mods for '{GetOtherGame()}' were skipped");
```
PrimeSonic: Should this logging be left to another class?

QModManager/API/ModLoading/Internal/QMod.cs
```diff
@@ -0,0 +1,339 @@
...
+    using System.Collections.Generic;
+    using System.IO;
+    using System.Linq;
+    using System.Reflection;
+    using Oculus.Newtonsoft.Json;
+    using QModManager.DataStructures;
+    using QModManager.Utility;
+
+    [JsonObject(MemberSerialization.OptIn)]
+    internal class QMod : IQMod, IQModSerialiable, IQModLoadable, ISortable<string>
```
PrimeSonic: There is a lot going on in this class:  initializing from JSON, from a DLL, as an invalid mod with a name only, and it includes the `QModManager` singleton.  

I would like to see about refactoring all the common code into an abstract `QModBase` class so as to better separate the methods used for loading the mod from these different sources.

Thoughts?

QModManager/DataStructures/SortedTree.cs
```diff
@@ -0,0 +1,217 @@
...
+        {
+            var list = new List<IdType>(this.NodeCount);
+
+            erroredIds = ClearErrorsCleanTree();
+            CreateFlatIndexList(Root, list);
+
+            return list;
+        }
+
+        private static void CreateFlatIndexList(SortedTreeNode<IdType, DataType> node, ICollection<IdType> list)
```
PrimeSonic: Returning an IndexList is redundant when the DataType is `ISortable<IdType>`, which already has the `Id` property.  

Will require updating some unit tests, but the "list of indexes" scenario isn't needed, so get rid of it.

QModManager/DataStructures/SortedTreeNode.cs
AlexejheroYTB: Is it really necessary?

QModManager/Patching/GameDetector.cs
```diff
@@ -0,0 +1,45 @@
...
+    {
+        internal readonly Game CurrentlyRunningGame;
+
+        internal bool IsValidGameRunning => CurrentlyRunningGame == Game.Subnautica || CurrentlyRunningGame == Game.BelowZero;
+
+        internal GameDetector()
+        {
+            bool isSubnautica = Directory.GetFiles(Environment.CurrentDirectory, "Subnautica.exe", SearchOption.TopDirectoryOnly).Length > 0
+                || Directory.GetFiles(Environment.CurrentDirectory, "Subnautica.app", SearchOption.TopDirectoryOnly).Length > 0
+                || Directory.GetDirectories(Environment.CurrentDirectory, "Subnautica.app", SearchOption.TopDirectoryOnly).Length > 0;
```
AlexejheroYTB: `.app` is a directory extension, not a file extension. The `GetFiles` call can be safely removed

QModManager/Patching/GameDetector.cs
```diff
@@ -0,0 +1,45 @@
...
+                CurrentlyRunningGame = Game.Subnautica;
+            }
+            else if (isBelowZero && !isSubnautica)
+            {
+                Logger.Info("Detected game: BelowZero");
+                CurrentlyRunningGame = Game.BelowZero;
+            }
+            else if (isSubnautica && isBelowZero)
+            {
+                Logger.Fatal("A fatal error has occurred.", "Both Subnautica and Below Zero files detected!");
```
AlexejheroYTB: For all fatal errors, block the game from running with a dialog with no buttons

QModManager/Patching/GameDetector.cs
```diff
@@ -0,0 +1,45 @@
...
+                CurrentlyRunningGame = Game.Both;
+            }
+            else
+            {
+                Logger.Fatal("A fatal error has occurred.", "No game executable was found!");
```
AlexejheroYTB: Same here

QModManager/DataStructures/SortedTreeNode.cs
PrimeSonic: I look at `SortBefore(other, true)` and it doesn't make sense what makes it different from `SortBefore(other)`.

I'd prefer it to be more readable, even if it's functionally identical.

QModManager/API/Game.cs
PrimeSonic: Main issue is that `Game` is a class in the Unity assemblies.  

My only thought would be to use something like `QModGame`.

QModManager/API/ModLoading/Internal/Initializer.cs
PrimeSonic: So far, it seems like each class is doing its own logging, but logging might be something we can centralize, leaving these classes to be purely functional.  
I'll be looking into it later.

QModManager/DataStructures/SortResults.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace QModManager.DataStructures
+{
+    internal enum SortResults : int
+    {
+        NoSortPreference = 0, // Okay
+        SortBefore = 1, // Okay
+        SortAfter = 2, // Okay
+        CircularLoadOrder = 3, // Error
+        CircularDependency = 6, // Error
+        DuplicateId = 12, // Error
```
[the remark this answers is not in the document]
PrimeSonic: There was a method to this madness, but I will double-check if these numbers can't be make into something more intuitive, or if `[Flags]` wouldn't be a better fit.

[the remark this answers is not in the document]
AlexejheroYTB: Why... would this need to be a `[Flags]` enum? There's no point in doing this

QModManager/API/ModLoading/Internal/Initializer.cs
```diff
@@ -0,0 +1,178 @@
...
+        }
+
+        private static void PatchSMLHelper()
+        {
+            Logger.Info($"Loading SMLHelper...");
+            try
+            {
+                HarmonyInstance Harmony = Patcher.Harmony;
+
+                CustomFishPatcher.Patch(Harmony);
```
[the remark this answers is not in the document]
PrimeSonic: I guess so. It's a static field, not a property. Shouldn't be any less efficient to call it directly.

QModManager/API/ModLoading/Internal/PatchMethod.cs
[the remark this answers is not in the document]
PrimeSonic: This is mostly a rework of the original Patcher invocation from master.  
https://github.com/QModManager/QModManager/blob/master/QModManager/Patcher.cs#L291

Are you sure this isn't correctly?  
I have unit tests showcasing this code being run.  
https://github.com/QModManager/QModManager/blob/dev/ModLoading/Unit%20Tests/QModManager/InitializerTests.cs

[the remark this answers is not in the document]
AlexejheroYTB: If the patch method is static the first argument is ignored. This is qwiso's mistake

QModManager/API/ModLoading/Internal/PatchingOrder.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace QModManager.API.ModLoading.Internal
+{
+    /// <summary>
+    /// FOR INTERNAL USE ONLY.
+    /// </summary>
+    public enum PatchingOrder
```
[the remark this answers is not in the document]
PrimeSonic: This one is `public` because it's being used by the [IPatchOrder](https://github.com/QModManager/QModManager/blob/dev/ModLoading/QModManager/API/ModLoading/Internal/IPatchMethod.cs) interface.  
And interface method implementations need to be public.  
I know it's not ideal to have this exposed, but there really isn't anything an external assembly could do with this.  
The various [QModdPatch](https://github.com/QModManager/QModManager/blob/dev/ModLoading/QModManager/API/ModLoading/QModPatchMethod.cs) methods don't even allow this to be changed.

QModManager/API/ModLoading/Internal/QMod.cs
```diff
@@ -0,0 +1,339 @@
...
+    {
+        /// <summary>
+        /// The dummy <see cref="QMod"/> which is used to represent QModManager
+        /// </summary>
+        internal static QMod QModManager { get; } = new QMod
+        {
+            Id = "QModManager",
+            DisplayName = "QModManager",
+            Author = "QModManager Dev Team",
```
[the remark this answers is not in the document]
PrimeSonic: Yeah, that wording is bad and should be updated.  
I could agree to "The QModManager dev team".

QModManager/API/ModLoading/Internal/PatchingOrder.cs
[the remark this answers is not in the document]
AlexejheroYTB: I do not think that there is a need for the `IPatchMethod` interface.

QModManager/API/ModLoading/Internal/QMod.cs
```diff
@@ -0,0 +1,339 @@
...
+        {
+            switch (this.Game)
+            {
+                case "BelowZero":
+                    _moddedGame = API.Game.BelowZero;
+                    break;
+                case "Both":
+                    _moddedGame = API.Game.Both;
+                    break;
+                case "Subnautica":
```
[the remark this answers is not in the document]
PrimeSonic: The optimizer will remove it at compile time. This is here as documentation.

Alternatively, this should probably instead be doing more stricture checking to make sure that the provided text is correct instead of just defaulting to "Subnautica" when an invalid string is provided.

QModManager/API/QModAPI.cs
```diff
@@ -1,4 +1,7 @@
-﻿using QModManager.Utility;
+﻿using QModManager.API.ModLoading;
+using QModManager.API.ModLoading.Internal;
+using QModManager.Utility;
+using System;
```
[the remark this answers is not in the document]
PrimeSonic: Will check what happened here.  
If this happened because `global::` tags showed up in the usings, that may have been why.  
Otherwise, this is just cleanup to do.

QModManager/DataStructures/ErrorTypes.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace QModManager.DataStructures
```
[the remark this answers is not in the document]
PrimeSonic: I'm really not a fan of spaces inside folder or file names, but I'll live with this one.

QModManager/DataStructures/SortResults.cs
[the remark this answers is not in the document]
PrimeSonic: It's part of the code that checks for circular load orders and circular dependencies.  
https://github.com/QModManager/QModManager/blob/dev/ModLoading/QModManager/DataStructures/SortedTreeNode.cs#L198  
Again, I'll double-check and see if this needs explicit values at all.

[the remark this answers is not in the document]
AlexejheroYTB: A `[Flags]` enum means that a variable can have multiple values at the same time, and the numbering is 1, 2, 4, 8, 16, 32, etc.

QModManager/API/ModLoading/Internal/PatchingOrder.cs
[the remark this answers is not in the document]
PrimeSonic: I see a way to do this without that interface.  
Yes, this public enum can be made internal with just a little rework of the patch method attributes.

QModManager/API/ModLoading/Internal/PatchMethod.cs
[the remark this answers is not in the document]
PrimeSonic: > This is qwiso's mistake

lol, wow.  
Okay, let's do this correctly then.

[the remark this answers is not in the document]
PrimeSonic: These methods _shouldn't_ be requiring parameters in the first place.  
But yes, we can check this for safety to avoid an unnecessary exception catch.

QModManager/API/ModLoading/Internal/QMod.cs
```diff
@@ -0,0 +1,339 @@
...
+        }
+
+        private Dictionary<PatchingOrder, PatchMethod> GetPatchMethods(Type originatingType)
+        {
+            var dictionary = new Dictionary<PatchingOrder, PatchMethod>(3);
+
+            MethodInfo[] methods = originatingType.GetMethods(BindingFlags.Public);
+            foreach (MethodInfo method in methods)
+            {
+                var patchMethods = (QModPatchMethod[])method.GetCustomAttributes(typeof(QModPatchMethod), false);
```
PrimeSonic: Make corrections here relating to the changes to the `IPatchMethod`.

QModManager/API/ModLoading/Internal/QModFactory.cs
```diff
@@ -0,0 +1,211 @@
...
+    using QModManager.DataStructures;
+    using QModManager.Utility;
+
+    internal class QModFactory
+    {
+        private SortedTree<string, QMod> sortedMods;
+        private List<QMod> unloadedMods;
+        private string[] subDirectories;
+
+        internal int FailedToCreate { get; private set; }
```
PrimeSonic: Should probably be an `out` parameter of the `BuildModLoadingList` method.

QModManager/API/ModLoading/QModPatchMethod.cs
```diff
@@ -0,0 +1,26 @@
...
+        /// Initializes a new instance of the <see cref="QModPatchMethod"/> class for normal patching.        
+        /// </summary>
+        public QModPatchMethod()
+        {
+        }
+
+        /// <summary>
+        /// Gets the patch method's meta patch order.
+        /// </summary>
+        public PatchingOrder PatchOrder { get; } = PatchingOrder.NormalInitialize;
```
PrimeSonic: Remove `IPatchMethod` if possible and make this property `internal`.

QModManager/API/ModLoading/IQMod.cs
PrimeSonic: Note that `.NET 3.5` does have `IReadOnlyCollection` or `IReadOnlyList`.  
So the best we can do here is an `IEnumerable`.

QModManager/API/ModLoading/Internal/QModFactory.cs
```diff
@@ -0,0 +1,211 @@
...
+                    Console.WriteLine($"- {mod.DisplayName} ({mod.Id})");
+                }
+            }
+        }
+
+        private static QMod FromDll(string subDirectory, string[] dllFilePaths)
+        {
+            foreach (string dllFile in dllFilePaths)
+            {
+                var assembly = Assembly.LoadFrom(dllFile);
```
PrimeSonic: @AlexejheroYTB 
Is there potential issue if we load an assembly that wasn't needed?  
ex. An extra DLL included by a mod that it uses.

Examples/PatchExample.cs
```diff
@@ -0,0 +1,19 @@
...
+    [QModLoadBefore("SomeMod")]
+    [QModLoadBefore("SomeMod2")]
+    [QModLoadAfter("SomeOtherMod")]
+    [QModDependency("ApiMod")]
+    [QModDependency("OtherAPIMod", 2, 2, 1)]
+    internal static class PatchExample
+    {
+        [QModPatchMethod]
+        public static void PatchMyMod()
+        {
```
AlexejheroYTB: Idea: Let patch methods return a `State` which can either be `OK`, `Error`, or `Cancel` (names up for debate) that change the way QModManager treats them. For example, if `State.Error` is returned, the mod is marked as errored, and if `State.Cancel` is returned, the mod's loading is cancelled but it's not mark as errored.

PrimeSonic: Not a terrible idea. But we can't enforce anything about method signatures with attributes.  
Are you still sure you want to include this?

QModManager/API/ModLoading/Internal/ModLoadingResults.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace QModManager.API.ModLoading.Internal
+{
+    internal enum ModLoadingResults
+    {
+        Success,
+        NoMethodToExecute,
+        Failure,
+        AlreadyLoaded,
+        CurrentGameNotSupported,
```
PrimeSonic: @AlexejheroYTB As either a replacement or supplement to the current status enumns in this branch,   I'm proposing this `ModStatus` set to be used across the entirety of the mod loading life cycle.  
Once all loading is complete, we would have a final collection of all mods and their final status to report on in a single block.
```CSharp
internal enum ModStatus
{
  Success = 0,
  CanceledByAuthor = 1,
  PatchMethodFailed = 2,
  TooManyPatchMethods = 1,
  MissingPatchMethod = 2,
  CircularLoadOrder = 3,
  CircularDependency = 4,
  MissingDependency = 5,
  CurrentGameNotSupported = 6,
  DuplicateIdDetected = 7,
  MissingCoreData = 8,
  MissingAssemblyFile = 9
}
```
The values define how early the error happens in the loading life-cycle.  
Higher values are more critical.  
If we have a `Dictionary<QMod, ModStatus> LoadingStatus` somewhere, then we can always update it with something like this
```CSharp
LoadingStatus[mod] = Math.Max(LoadingStatus[mod], newStatus);
```
Which will keep the relevant status on top no matter what else we do with the QMod instance along the way.

QModManager/API/ModLoading/Internal/PatchMethod.cs
PrimeSonic: Non-static method handling still pending

QModManager/API/ModLoading/Internal/QMod.cs
PrimeSonic: Pending

Examples/PatchExample.cs
```diff
@@ -0,0 +1,32 @@
...
+    [QModLoadAfter("SomeOtherMod")]
+    [QModDependency("ApiMod")]
+    [QModDependency("OtherAPIMod", 2, 2, 1)]
+    internal static class PatchExample
+    {
+        [QModPrePatchMethod]
+        public static PatchResults PatchMyModEarly()
+        {
+            // Early mod patching happens here
+            return PatchResults.OK;
```
AlexejheroYTB: Is `PatchResults` only available for `[QModPrePatchMethod]`?

QModManager/API/ModLoading/Internal/QModFactory.cs
```diff
@@ -0,0 +1,228 @@
...
+                    continue;
+
+                QMod mod;
+
+                string jsonFile = Path.Combine(subDir, "mod.json");
+
+                if (File.Exists(jsonFile))
+                    mod = FromJsonFile(subDir);
+                else
+                    mod = FromDll(subDir, dllFiles);
```
AlexejheroYTB: First try to load from DLL, then fallback to json file.

QModManager/API/ModLoading/Internal/QModFactory.cs
```diff
@@ -0,0 +1,228 @@
...
+                        throw new InvalidOperationException("Invalid error reported by mod sorter");
+                }
+            }
+
+            CheckOldHarmony(modsToLoad);
+
+            return modList;
+        }
+
+        //private static int ReportErrors(List<string> erroredMods)
```
AlexejheroYTB: What's happening with this?

QModManager/API/ModLoading/Internal/QModFactory.cs
AlexejheroYTB: Not really a problem. Although we should probably load it into a new `AppDomain` (`AppDomain.CreateDomain`) and then either load it into the main domain if it's a mod dll or unload it by unloading the domain.

QModManager/API/ModLoading/Internal/QModFactory.cs
PrimeSonic: Since I'm moving this towards a single list containing all mods paired with their results,  
I plan to create a specialized logging class that can be provided with this list, at the various stages of patching, to handle logging of errors and ultimately if the Dialog pop-up should be used or not.

Examples/PatchExample.cs
PrimeSonic: No. There is no restriction in place for that.  
It's just difficult to show an optional return value with a single example.

AlexejheroYTB: People should know that this works with all methods. Maybe a comment?

QModManager/API/ModLoading/Internal/PatchMethod.cs
AlexejheroYTB:
```csharp
var instance = Activator.CreateInstance(this.Method.DeclaringType);
this.Method.Invoke(instance, null);
```

QModManager/API/ModLoading/Internal/QModPlaceholder.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        private readonly ModStatus status;
+        /// <summary>
+        /// The dummy <see cref="QMod"/> which is used to represent QModManager
+        /// </summary>
+        internal static QModPlaceholder QModManager { get; } = new QModPlaceholder
+        {
+            Id = "QModManager",
+            DisplayName = "QModManager",
+            Author = "The QModManager Dev Team",
```
AlexejheroYTB: `the QModManager dev team` or `The QModManager Dev Team`?

QModManager/DataStructures/ErrorTypes.cs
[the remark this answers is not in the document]
AlexejheroYTB: Actually, nevermind. Keep it this way.

QModManager/API/ModLoading/Internal/QModPlaceholder.cs
PrimeSonic: I'm partial to **T**itle **C**asing

AlexejheroYTB: Then also change this in `AssemblyInfo.cs` (for all of the projects) and in `QModsInstallerScript.iss` please

QModManager/API/ModLoading/Internal/QModFactory.cs
```diff
@@ -0,0 +1,219 @@
...
+                    {
+                        modsThatUseOldHarmony.Add(mod);
+                        break;
+                    }
+                }
+            }
+
+            if (modsThatUseOldHarmony.Count > 0)
+            {
+                Logger.Warn($"Some mods are using an old version of harmony! This will NOT cause any problems, but it's not recommended:");
```
AlexejheroYTB: This could be merged with other logs

QModManager/API/ModLoading/Internal/QModPlaceholder.cs
AlexejheroYTB: Or just change this to `QModManager`. Easier that way

QModManager/API/ModLoading/Internal/QMod.cs
```diff
@@ -0,0 +1,134 @@
...
+
+                case PatchResults.ModderCanceled:
+                    this.PatchMethods.Clear(); // Do not attempt any other patch methods
+                    return ModLoadingResults.CancledByModAuthor;
+            }
+
+            return ModLoadingResults.Failure;
+        }
+
+        protected static bool IsDefaultVersion(Version version)
```
PrimeSonic: TODO: Handle version requirements for dependencies

QModManager/API/ModLoading/Internal/QModFactory.cs
AlexejheroYTB: This would most likely require adding a [new dependency](https://github.com/jduv/AppDomainToolkit)

QModManager/API/ModLoading/Internal/QPatchMethod.cs
```diff
@@ -0,0 +1,62 @@
...
+        internal PatchingOrder Order { get; }
+        internal MethodInfo Method { get; }
+        internal Assembly LoadedAssembly { get; }
+        internal bool IsPatched { get; private set; }
+
+        internal PatchResults TryInvoke()
+        {
+            try
+            {
+                // TODO - Add handling for non-static methods
```
AlexejheroYTB: TODO - Add handling for non-static methods

QModManager/API/ModLoading/Internal/QModPlaceholder.cs
```diff
@@ -0,0 +1,41 @@
...
+            status = ModStatus.Success;
+        }
+
+        internal QModPlaceholder(string name)
+        {
+            this.Id = Patcher.IDRegex.Replace(name, "");
+            this.DisplayName = name;
+            this.Author = "Unknown";
+            this.SupportedGame = QModGame.None;
+            this.Enable = false;
```
AlexejheroYTB: How does `Enable` work currently?

QModManager/API/QModAPI.cs
```diff
@@ -129,12 +132,13 @@ ReadOnlyCollection<IQMod> IQModAPI.GetAllMods(bool includeUnloaded)
         /// <returns>A read only list of mods containing all of the loaded mods, and optionally unloaded/errored mods</returns>
         ReadOnlyCollection<IQMod> IQModAPI.GetAllMods(bool includeUnloaded, bool includeErrored)
         {
-            if (includeErrored)
-                return Patcher.foundMods.Select(qmod => (IQMod)qmod).ToList().AsReadOnly();
-            else if (includeUnloaded)
-                return Patcher.sortedMods.Select(qmod => (IQMod)qmod).ToList().AsReadOnly();
-            else
-                return Patcher.loadedMods.Select(qmod => (IQMod)qmod).ToList().AsReadOnly();
+            throw new NotImplementedException();
```
AlexejheroYTB: TODO - QModAPI

QModManager/Dialog.cs
```diff
@@ -6,7 +6,6 @@
 using System.Diagnostics;
```
AlexejheroYTB: This class could have a better name and could be moved someplace else, maybe in `Utility`?

QModManager/Patching/GameDetector.cs
```diff
@@ -0,0 +1,43 @@
...
+                CurrentlyRunningGame = QModGame.Subnautica;
+            }
+            else if (isBelowZero && !isSubnautica)
+            {
+                Logger.Info("Detected game: BelowZero");
+                CurrentlyRunningGame = QModGame.BelowZero;
+            }
+            else if (isSubnautica && isBelowZero)
+            {
+                Logger.Fatal("A fatal error has occurred.", "Both Subnautica and Below Zero files detected!");
```
AlexejheroYTB: The log created by this is not very pretty

```
[QModManager] [GameDetector] [Fatal] A fatal error has occurred.
Both Subnautica and Below Zero files detected!
```

QModManager/Patching/Patcher.cs
```diff
@@ -0,0 +1,163 @@
+namespace QModManager
```
AlexejheroYTB: Please keep folder structure and namespaces consistent

QModManager/SummaryLogger.cs
```diff
@@ -0,0 +1,70 @@
...
+
+            foreach (QMod mod in mods)
+            {
+                if (mod.IsLoaded && mod.HarmonyOutdated)
+                    modsThatUseOldHarmony.Add(mod);
+            }
+
+            if (modsThatUseOldHarmony.Count > 0)
+            {
+                Logger.Warn($"Some mods are using an old version of harmony! This will NOT cause any problems, but it's not recommended:");
```
AlexejheroYTB: Move this to `LogSummaries`

QModManager/API/ModLoading/Internal/PatchMethod.cs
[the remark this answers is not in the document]
AlexejheroYTB: I think that this should be done in the attribute class itself.

QModManager/Dialog.cs
PrimeSonic: Utility seems like the perfect place for this. 👍

QModManager/API/ModLoading/Internal/QModFactory.cs
```diff
@@ -0,0 +1,167 @@
...
+                        modList.Add(erroredMod.Key, ModStatus.CircularDependency);
+                        break;
+                    case ErrorTypes.CircularLoadOrder:
+                        modList.Add(erroredMod.Key, ModStatus.CircularLoadOrder);
+                        break;
+                    case ErrorTypes.MissingDepency:
+                        modList.Add(erroredMod.Key, ModStatus.MissingDependency);
+                        break;
+                    default:
+                        throw new InvalidOperationException("Invalid error reported by mod sorter");
```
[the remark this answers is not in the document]
PrimeSonic: This one here should be changed into a `FalatPatchingException`.  
In this case in particular, hitting this `default` cause should absolutely never happen.

QModManager/API/ModLoading/Internal/QModPlaceholder.cs
PrimeSonic: Handled in `QModFactory`, this in intended to prevent the mod from patching.
```CSharp
if (!mod.Enable)
{
    earlyErrors.Add(mod, ModStatus.CanceledByUser);
    continue;
}
```
As for how new mods will change this value, that is still yet to be determined unless we leave it inside the `mod.json` and keep using that file (in a simplified form) to handle things like this.

QModManager/API/ModLoading/Internal/Initializer.cs
```diff
@@ -0,0 +1,124 @@
...
+                return true; // Test mode
+
+            return UpdateSMLHelper() && PatchSMLHelper();
+        }
+
+        private void InitializeMods(PairedList<QMod, ModStatus> modsToInitialize, PatchingOrder order)
+        {
+            foreach (Pair<QMod, ModStatus> pair in modsToInitialize)
+            {
+                QMod mod = pair.Key;
```
PrimeSonic: TODO: Skip any mods where ModStatus is not `ModStatus.Success`.

QModManager/API/ModLoading/Internal/QModFactory.cs
PrimeSonic: Heads up. I'm going to need your help for this one.  
Dynamically loading DLLs at runtime isn't something I've worked with before.

QModManager/API/ModLoading/Internal/PatchMethod.cs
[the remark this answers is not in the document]
PrimeSonic: Are you suggesting we merge the functionality of `QModPatchAttributeBase` and `QModPatchMethod` into a single class?  
I think it's doable. I'm on the fence about it right now but I don't have any strong reasons not to do it if you think it'll make sense to make it happen.

QModManager/API/ModLoading/Internal/QModFactory.cs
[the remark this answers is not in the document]
AlexejheroYTB: Can exceptions other than FatalErrorException (or whatever its called) reach the main patch method?

QModManager/API/ModLoading/Internal/QModFactory.cs
AlexejheroYTB: Important question: do we want to unload unnecessary assemblies after QMod attribute was found?

QModManager/API/ModLoading/Internal/QModFactory.cs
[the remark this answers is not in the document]
PrimeSonic: They shouldn't unless we did something wrong. 
Any other type of exception should be getting caught and handled as close to the source as possible. 
While these are the exceptions we will _deliberately_ throw to be caught by the main patcher, any exceptions caught at that high level should halt the patching process.

QModManager/SummaryLogger.cs
PrimeSonic: This method is already called from the `LogSummaries` method.  
Did you have something else in mind?

AlexejheroYTB: So far all of the calls to `LogStatus` are for errored/cancelled mods, that is not where _warn_ logging is done.

QModManager/SummaryLogger.cs
```diff
@@ -0,0 +1,70 @@
...
+        {
+            CheckOldHarmony(mods.Keys);
+            LogStatus(mods, ModStatus.CanceledByAuthor, "The following mods have been disabled internally by the mod author", Logger.Level.Info);
+            LogStatus(mods, ModStatus.CanceledByUser, "The following mods have been disabled by user configuration", Logger.Level.Info);
+            LogStatus(mods, ModStatus.PatchMethodFailed, "The following mods failed during patching", Logger.Level.Error);
+            LogStatus(mods, ModStatus.TooManyPatchMethods, "The following mods had too many patch methods and were canceled", Logger.Level.Warn);
+            LogStatus(mods, ModStatus.MissingPatchMethod, "The following mods had no patch methods to run", Logger.Level.Warn);
+            LogStatus(mods, ModStatus.CircularLoadOrder, "The following mods could not be loaded due to circular load order", Logger.Level.Warn);
+            LogStatus(mods, ModStatus.CircularDependency, "The following mods could not be loaded due to circular dependencies", Logger.Level.Warn);
+            LogStatus(mods, ModStatus.MissingDependency, "The following mods could not be loaded due to missing dependencies", Logger.Level.Warn);
```
AlexejheroYTB: Everything is here should use `Logger.Level.Error`

PrimeSonic: I've used `Warn` for all issues that were caught _before_ we started any patching.  
These shouldn't cause any major problems in-game since the mods were never loaded.  

I used `Error` for the issues that come up during the patching which could leave the game in a half-patched state.

QModManager/SummaryLogger.cs
PrimeSonic: True. But I don't see any better place to put this, unless it's moved somewhere else entirely.

QModManager/API/ModLoading/QModPatchMethod.cs
```diff
@@ -5,6 +5,7 @@
 
     /// <summary>
     /// Identifies a normal patch method for a QMod.<para/>
+    /// This method must be public, must take no parameters, and must return either <seealso cref="void"/> or <seealso cref="PatchResults"/>.<para/>
```
AlexejheroYTB: Doesnt really have to be public

PrimeSonic: While it would be possible to make this work without it being `public`, it would be "more correct" if it were.  
This is one assembly exposing methods to be invoked by another, external assembly.  
When that is the intent, you make the method and class `public`.

Having the method be `public` is no different to how mod patching methods are currently being done.

AlexejheroYTB: Doesnt it already work for non-public methods?

PrimeSonic: I can't say for sure as I don't have any examples of a mod having these methods as non-public.  
Even if it does work without the method being public, there's no harm in asking modders to keep them public for simplicity.

QModManager/SummaryLogger.cs
AlexejheroYTB: From an user's perspective, mods that failed to load are "errored"

QModManager/API/ModLoading/Internal/Initializer.cs
```diff
@@ -0,0 +1,127 @@
+﻿namespace QModManager.API.ModLoading
```
AlexejheroYTB: The `Internal` folder inside of `QModManager/API/ModLoading` should be moved someplace else.

QModManager/Patching/Patcher.cs
```diff
@@ -0,0 +1,171 @@
...
+
+                Logger.Info($"Loading QModManager v{Assembly.GetExecutingAssembly().GetName().Version.ToStringParsed()}...");
+
+                if (QModBaseDir == null)
+                {
+                    Logger.Fatal("A fatal error has occurred.");
+                    Logger.Fatal("There was an error with the QMods directory");
+                    Logger.Fatal("Please make sure that you ran Subnautica from Steam/Epic/Discord, and not from the executable file!");
+
+                    Dialog.Show("A fatal error has occurred. QModManager could not be initialized.", Dialog.Button.SeeLog, Dialog.Button.Disabled, false);
```
AlexejheroYTB: In this case we don't want to prevent the game from running, since the user can just run it without QMods, so having the second button `Dialog.Button.Close` would make more sense.

QModManager/Patching/Patcher.cs
```diff
@@ -0,0 +1,171 @@
...
+                    else
+                        erroredMods++;
+                }
+
+                ErrorModCount = erroredMods;
+
+                Logger.Info($"Finished loading QModManager. Loaded {loadedMods} mods");
+
+                if (ErrorModCount > 0)
+                    Logger.Info($"A total of {ErrorModCount} mods failed to load");
```
AlexejheroYTB: `Logger.Warn`?

QModManager/Patching/Patcher.cs
```diff
@@ -0,0 +1,171 @@
...
+
+                SummaryLogger.LogSummaries(modsToLoad);
+            }
+            catch (FatalPatchingException pEx)
+            {
+                Logger.Fatal($"FATAL PATCHING EXCEPTION! - Patching ended");
+                Logger.Exception(pEx);
+
+                Dialog.Show("A fatal patching exception has been caught. QModManager could not be initialized.", Dialog.Button.SeeLog, Dialog.Button.Disabled, false);
```
AlexejheroYTB: Upon clicking any button, the dialog will go away, which is not what we want. `Dialog` will have to be updated to allow dialogs to remain on screen even if a button was clicked.

QModManager/API/QModAPI.cs
PrimeSonic: Pushing this out for 3.1

QModManager/API/ModLoading/Internal/QModPlaceholder.cs
PrimeSonic: So we might just continue using `mod.json` for this. 
`SAM` and `Vortex` are likely both using this field.

I'll have to make some adjustments to the DLL loading so that it still checks the `mod.json` for that value.

QModManager/API/QModAPI.cs
```diff
@@ -1,15 +1,17 @@
...
+using QModManager.Patching;
+using QModManager.Utility;
 
 namespace QModManager.API
 {
     /// <summary>
     /// The main class for the QModManager API
     /// </summary>
-    public partial class QModAPI : IQModAPI
+    public class QModAPI : IQModAPI
```
AlexejheroYTB: My opinion would be to get rid of this class entirely

QModManager/API/QModHooks.cs
```diff
@@ -54,11 +54,11 @@ internal static void Load()
         {
             SceneManager.sceneLoaded += (scene, loadSceneMode) => SceneLoaded?.Invoke(scene, loadSceneMode);
 
-#pragma warning disable CS0618 // Type or member is obsolete
-            Hooks.Load();
+//#pragma warning disable CS0618 // Type or member is obsolete
+//            Hooks.Load();
```
AlexejheroYTB: Why disable hooks?

PrimeSonic: After removing the Mod API classes these references were broken too. 
Commented then out just to get things compiling again.

QModManager/Patching/QModFactory.cs
```diff
@@ -0,0 +1,214 @@
...
+            var earlyErrors = new PairedList<QMod, ModStatus>(subDirectories.Length);
+
+            foreach (string subDir in subDirectories)
+            {
+                string[] dllFiles = Directory.GetFiles(subDir, "*.dll", SearchOption.TopDirectoryOnly);
+
+                if (dllFiles.Length < 1)
+                    continue;
+
+                QMod mod = FromDll(subDir, dllFiles);
```
PrimeSonic: @AlexejheroYTB I never actually got this working.  
If there is any chance for QMM 3.0 to move forward, we need to rip out the DLL based loading and continue with `mod.json` and/or some other manifest file.

Everything else is still an improvement to mod loading, especially with the 3 patch methods, but this DLL stuff needs to go.

Same goes for SMLHelper.  We will need to re-think (and re-do) the integration with proper consideration for backwards compatibility.

QModManager/Patching/Patcher.cs
```diff
@@ -55,7 +55,7 @@ internal static void Patch()
                     Logger.Fatal("There was an error with the QMods directory");
                     Logger.Fatal("Please make sure that you ran Subnautica from Steam/Epic/Discord, and not from the executable file!");
 
-                    Dialog.Show("A fatal error has occurred. QModManager could not be initialized.", Dialog.Button.SeeLog, Dialog.Button.Disabled, false, true);
+                    Dialog.Show("A fatal error has occurred. QModManager could not be initialized.", Dialog.Button.Close, Dialog.Button.Disabled, false, true);
```
AlexejheroYTB: The last parameter of this function is `cannotClose`. If this parameter is set to true, the dialog cannot be closed, so a button of `Close` doesn't make sense here.

QModManager/OptionsManager.cs
```diff
@@ -1,36 +1,36 @@
...
-                ModsTab = __instance.AddTab("Mods");
-                __instance.AddHeading(ModsTab, "QModManager");
+        //[HarmonyPatch(typeof(uGUI_OptionsPanel), "AddTabs")]
+        //private static class OptionsPatch
+        //{
+        //    [HarmonyPostfix]
+        //    private static void Postfix(uGUI_OptionsPanel __instance)
+        //    {
+        //        ModsTab = __instance.AddTab("Mods");
+        //        __instance.AddHeading(ModsTab, "QModManager");
```
AlexejheroYTB: Why did you delete this?

QModManager/QModManager.csproj
```diff
@@ -81,13 +81,11 @@
...
     </Reference>
-    <Reference Include="Assembly-CSharp-firstpass_publicized">
-      <HintPath>..\Dependencies\Assembly-CSharp-firstpass_publicized.dll</HintPath>
-      <Private>False</Private>
+    <Reference Include="Assembly-CSharp">
+      <HintPath>..\Dependencies\Assembly-CSharp.dll</HintPath>
     </Reference>
-    <Reference Include="Assembly-CSharp_publicized">
-      <HintPath>..\Dependencies\Assembly-CSharp_publicized.dll</HintPath>
-      <Private>False</Private>
```
AlexejheroYTB: Set these to `<Private>False</Private>` so they don't show up in `Build`

QModManager/Utility/Dialog.cs
```diff
@@ -20,30 +20,30 @@ internal class Button
...
+
+            //    if (File.Exists(logPath))
+            //    {
+            //        Process.Start(logPath);
+            //    }
+            //    else
+            //    {
+            //        Logger.Error($"Expected log file at: \"{logPath}\" but none was found.");
+            //    }
+            //});
```
AlexejheroYTB: Why was this removed?

QModManager/OptionsManager.cs
PrimeSonic: likely a mistake from trying to get the solution building again

QModManager/Utility/Dialog.cs
PrimeSonic: Because the dialog popup was throwing exceptions itself whenever it had to pop up, 
and the log file is no longer in this location.

I would ask you to fix this part since you know it better than I do.

QModManager/Patching/Patcher.cs
PrimeSonic: Everything to do with the dialog popup I leave to you

PR #105, opened 2019-06-18
Add project assets

AlexejheroYTB:


AlexejheroYTB:
Any other assets that would be useful to have in the repo?

kylinator25:
Do we have somewhere we can put misc files that would be useful to mod authors? Is this just for QMod specific assets?

AlexejheroYTB:
Explaing `misc files` please.

kylinator25:
Stuff like ingame prefab lists, just to put it all in an accessible place rather than modders having to dig through game files

AlexejheroYTB:
This pull request is just about QModManager specific assets.

PR #117, opened 2019-12-25
QModManager 3.0 Overhaul

PrimeSonic:
# QMM 3.0
Massive overhaul to the internals of the mod loading sequence plus additional features to improve compatibility between mods that are sensitive to load order.

---

### Attention Beta Testers
**Download the 3.0 Release Candidate [here](https://github.com/SubnauticaModding/QModManager/raw/devV3/Build/QModManager_Setup.exe)**

---

#### `Patcher` is no longer a single, monolithic class
Instead it is now more of an orchestrator class, delegating smaller tasks to specialized classes, improving readability and testing.

#### New data structures completely rework mod sorting and dependency checking
There is no longer a need for special handling of mods like SMLHelper. It will naturally be put at the bottom of the load order as it is sorted among all mods that depend on it.

#### Mods no longer required to include `EntryMethod` in their mod.json manifest
This can now be accomplished by adding the new custom attributes to the main patching class.  
`[QModCore]` identifies the main patching class.  
`[QModPatch]`, `[QModPrePatch]`, and `[QModPostPatch]` identify the main patching methods.

#### PrePatching and PostPatching is now possible
Mod initialization is now a 3 stage process;  
After sorting, all PrePatch methods will be invoked, followed by all Patch methods, and finally all PostPatch methods are invoked last.  
This allows greater control for mods that interact with other mods that may require specific timing to be able to guarantee that some patching code is run always before or always after normal mod initialization.  
This is feature only available via the new `[QModPrePatch]` and `[QModPostPatch]` attributes (wiki documentation in progress).

#### `VersionDependencies` are now simplified
You no longer need to specific operators like ">=" to inform that your mod requires a specific version "or later".  
Instead, all versions in version dependencies are now treated as the "_minimum version required_".  
As long as the required mod is at that version or later, the dependency check will pass.

#### Debug Logging no longer tied to registry
Debug logs for QMM are enabled simply by creating a file named "QModDebug.txt" in the Subnautica folder.

---

**WARNING:** As soon as this PR is merged into master, everyone launching the game will see that there is a newer version of QModManager available!

So make sure that QModManager has been released before the dev PR is merged


QModManager/API/ModLoading/QModPatchAttributeBase.cs
```diff
@@ -0,0 +1,99 @@
...
+
+            _secretPasword = secretPasword;
+        }
+
+        /// <summary>
+        /// Validates the that modder has read the documentation.
+        /// </summary>
+        /// <param name="method">The method.</param>
+        /// <exception cref="FatalPatchingException">This modder has not read the documentation and should not be using prepatch/postpatch functions.</exception>
+        internal void ValidateThatModderHasReadTheDocumentation(MethodInfo method)
```
PrimeSonic: **TODO** - Update this to _also_ use the QMod Id, as it may be different from the Assembly name and is guaranteed to be unique within the ordering while the Assembly name may not be.

QModManager/API/IQModServices.cs
```diff
@@ -0,0 +1,18 @@
...
+    /// </summary>
+    public interface IQModServices
+    {
+        /// <summary>
+        /// Gets a list successfully loaded mods.
+        /// </summary>
+        /// <returns>
+        /// The loaded mods.
+        /// </returns>
+        List<IQMod> GetLoadedMods();
```
PrimeSonic: Add the following service method:
```CS
IQMod FindMod(string id);  // Returns null if mod does not exist
```

AlexejheroYTB: With the notice that the method might not work correctly if used in PrePatch

PrimeSonic: > With the notice that the method might not work correctly if used in PrePatch

The method names and summaries were changed and no longer imply "successfully loaded". 

Also, this list is populated _before_ PrePatch methods are invoked.  
There is no way for this to fail, even on the earliest PrePatch.

Consumers of this service can check the `IsLoaded` get property of an `IQMod` to check if it has loaded or not.

PR #122, opened 2020-01-08
Merge mod: Quit To Desktop by RandyKnapp

AlexejheroYTB:
This pull request adds the (improved) features of the `Quit To Desktop` mod by @RandyKnapp into QModManager


QModManager/HarmonyPatches/QuitToDesktop.cs
```diff
@@ -0,0 +1,102 @@
+﻿// Original mod by RandyKnapp
+namespace QModManager.HarmonyPatches.QuitToDesktop
+{
+    using Harmony;
+    using UnityEngine;
+    using UnityEngine.UI;
+
+    internal static class QTD
```
PrimeSonic: It looks like the `QTD` class and the `IngameMenu_Start_Patch` class could be merged

QModManager/HarmonyPatches/QuitToDesktop.cs
```diff
@@ -0,0 +1,102 @@
...
+                yield return original.Current;
+
+                if (quitToDesktop && original.Current is IEnumerator)
+                    yield return new WaitForSecondsRealtime(2);
+
+                n++;
+            }
+        }
+    }
+    */
```
PrimeSonic: If this is no longer needed, remove the commented code.

QModManager/HarmonyPatches/QuitToDesktop.cs
```diff
@@ -0,0 +1,102 @@
+﻿// Original mod by RandyKnapp
```
PrimeSonic: Include this link in the comments 
`https://github.com/RandyKnapp/SubnauticaModSystem/tree/master/SubnauticaModSystem/QuitToDesktop`

QModManager/HarmonyPatches/QuitToDesktop.cs
AlexejheroYTB: `QTD` is also used in `IngameMenu_Open_Patch`

QModManager/HarmonyPatches/QuitToDesktop.cs
PrimeSonic: Is this commented code necessary?

AlexejheroYTB: It's my attempt to make QTD work in hardcore. It's 99% successful. I'll still work on it before 3.1 releases

MrPurple6411:
Ok so functionality seems to work fine. HOWEVER if we are going to have a prompt on the quit to desktop can it be a prompt to ask to save or not.... I mean I KNOW I pressed quit to desktop its not very easy to accidentally press quit. Also if someone says yes to save have a check to not quit if saving fails.

AlexejheroYTB:
The prompt is exactly the same as the one from the default quit button. I don't think I'll make it a save prompt.

MrPurple6411:
Ok. Just noting for the record that I am not a fan of an extra popup on QTD but if we must.

AlexejheroYTB:
It's so that people dont accidentally click it end erase all the progress they made in the last 2 hours because they forgot to save

MrPurple6411:
then making it a save prompt just makes sense..... ALSO how do you ACCIDENTALLY press it.. if you have clicked it you were at the least trying to go back to main menu.

AlexejheroYTB:
Exactly the same argument could be provided for the Quit to Main Menu button.

AlexejheroYTB:
For modders eager to exit as fast as possible, there's always ALT+F4

MrPurple6411:
yup and having a yes, no, cancel button on both. yes for save no for dont save and cancel means OOPS.

MrPurple6411:
so your argument is that the button is useless just use alt + f4

QModManager/HarmonyPatches/QuitToDesktop.cs
```diff
@@ -0,0 +1,120 @@
...
+    /*
+    [HarmonyPatch(typeof(IngameMenu), nameof(IngameMenu.QuitGame))]
+    internal static class IngameMenu_QuitGame
+    {
+        // This is a pass-through patch
+
+        [HarmonyPrefix]
+        internal static bool Prefix(IngameMenu __instance, bool quitToDesktop)
+        {
+            __instance.StartCoroutine(IngameMenu_QuitGameAsync.Passthrough(__instance.QuitGameAsync(quitToDesktop), quitToDesktop));
```
PrimeSonic: If you're having to make so many changes to the original mod,  
then you evidently don't think the mod in its original form is something that everyone should and would choose to use on their own.

So here you're trying to find a variation of this mod that would be good for everyone  
but in having to go through all that, you're pretty much giving all the evidence needed to say that this mod _isn't_ something that should be merged into the basic mod loader.

You are merging in a mod, but then it is different from the original,  
so people who actually want the original QtD in it's original form, won't be able to use it as it was
and saying "they can just Alt+F4" isn't a good answer.

My review is this:  
If QtD can't be merged without all these alterations that change how it originally worked,  
then it shouldn't be merged at all.

AlexejheroYTB: A Quit to Desktop button is something that all games should have. The only change I made to the mod was that I made the qtmm "are you sure" dialog also appear appear for the qtd button.

PrimeSonic: Then you aren't merging in the same mod anymore.

If you want to have this an an option, ask Randy if you can create a variant of QtD or submit to PR to Randy's mod that makes that optional.

Either way, this isn't the same mod anymore and you are going to be removing the option for people to use old behavior from people who might prefer it.

That is why my vote is still 'no' on merging this mod.

AlexejheroYTB: As an idea, I understand what you mean. In this case, though, the "old behavior" is one less click.

PrimeSonic: > As an idea, I understand what you mean. In this case, though, the "old behavior" is one less click.

You'd be surprised what people will complain about.

PR #125, opened 2020-01-09
Store QModManager configuration in qmodmanager-config.json

AlexejheroYTB:
- Added and Synchronized the disable console settings from the F3 debug menu to the mod options page
- Created a `Config` class which stores all configuration in `qmodmanager-config.json` in the root folder
- Deleted `PlayerPrefsExtra`


QModManager/Utility/Config.cs
```diff
@@ -0,0 +1,74 @@
...
+
+    internal static class Config
+    {
+        internal static class FIELDS
+        {
+            internal const string CHECK_FOR_UPDATES = "Check for updates";
+            internal const string ENABLE_CONSOLE = "Enable console";
+            internal const string ENABLE_DEBUG_LOGS = "Enable debug logs";
+            internal const string ENABLE_DEV_MODE = "Enable developer mode";
+        }
```
AlexejheroYTB: I don't know about this... @PrimeSonic?

PrimeSonic: Rather than expose constants and have to pass them in as parameters, instead, define an `interface` for the config and in there you can make specific `get` properties with the correct type and all the internal handling you'll need.

An example of this in action is in MSU
[IModConfig](https://github.com/PrimeSonic/PrimeSonicSubnauticaMods/blob/master/MoreCyclopsUpgrades/Config/IModConfig.cs])  
```CS
// The use of the interface keeps external call to the the config class clean
internal class ModConfig : IModConfig
{
    private static readonly ModConfig main = new ModConfig();

    // The singleton pattern allows you to use the interface from a static context
    internal static IModConfig Main
    {
        get
        {
            // If we need to load values from a file,
            // that can be done on demand
            if (!main.initialized)
                main.Initialize();

            return main;
        }
    }
```

AlexejheroYTB: I'd personally like to keep the Config class static rather than make it a singleton

PrimeSonic: Static classes can't implement interfaces.  
So with that in mind, your next best option is to define the static methods sort of like this:
```CS
internal static bool CheckForUpdates { get; private set; }
internal static bool EnableConsole { get; private set; }
internal static bool EnableDebugLogs { get; private set; }
internal static bool EnableDevMode { get; private set; }
```
You can (and probably should) hide away all of the logic around reading the config file from the rest of the assembly by keeping those methods private, and exposing only what external consumers actually care about: the final value.

PrimeSonic: You will probably end up making these as "get only" properties which contain the necessary code to read from the config file.

QModManager/Utility/Config.cs
```diff
@@ -7,33 +7,49 @@
...
+        internal static bool EnableConsole
+        {
+            get => Get("Enable console", false);
+            set => Set("Enable console", value);
+        }
+
+        internal static bool EnableDebugLogs
+        {
+            get => Get("Enable debug logs", false);
+            set => Get("Enable debug logs", value);
```
PrimeSonic: `Get` being called in the setter

QModManager/Utility/Config.cs
```diff
@@ -0,0 +1,90 @@
...
+
+                return null;
+            }
+        }
+
+        private static T Get<T>(string field, T def = default)
+        {
+            object value = Get(field);
+            if (value != null && value is T) return (T)value;
+            return def;
```
PrimeSonic: You can avoid redundant null checks using pattern matching.
```CS
if (value is T typedValue)
    return typedValue;

return def;
```

QModManager/Utility/Config.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        private static object Set(string field, object value, bool save = true)
+        {
+            if (!Loaded) Load();
+
+            Cfg.TryGetValue(field, out object oldValue);
+            Cfg.Remove(field);
+            Cfg.Add(field, value);
+            if (save) Save();
+            return oldValue;
```
PrimeSonic: You can remove the unused return value and make this method return `void`.

QModManager/Utility/Config.cs
```diff
@@ -0,0 +1,90 @@
...
+        internal static bool EnableDevMode
+        {
+            get => Get("Enable developer mode", false);
+            set => Set("Enable developer mode", value);
+        }
+
+        private static readonly string ConfigPath = Path.Combine(Environment.CurrentDirectory, "qmodmanager-config.json");
+
+        private static Dictionary<string, object> Cfg = new Dictionary<string, object>();
+        private static bool Loaded = false;
```
PrimeSonic: The value of the `Loaded` field is never changed.  
Every method that checks `if (Loaded)` is reloading the config every time.

QModManager/Utility/Config.cs
```diff
@@ -0,0 +1,90 @@
...
+            return def;
+        }
+
+        private static object Set(string field, object value, bool save = true)
+        {
+            if (!Loaded) Load();
+
+            Cfg.TryGetValue(field, out object oldValue);
+            Cfg.Remove(field);
+            Cfg.Add(field, value);
```
PrimeSonic:
```
The 3 lines used to set this can be replaced by one simple line:
```CS
Cfg[field] = value;
```

QModManager/Utility/Config.cs
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        private static T Get<T>(string field, T def = default)
+        {
+            object value = Get(field);
+            if (value != null && value is T) return (T)value;
+            return def;
+        }
+
+        private static object Set(string field, object value, bool save = true)
```
PrimeSonic: The default value of the optional `save` parameter is also redundant as it is only ever used as true.  
You can remove it and its usages and continue saving by default.

QModManager/Utility/Config.cs
```diff
@@ -0,0 +1,90 @@
...
+            Cfg = JsonConvert.DeserializeObject<Dictionary<string, object>>(text);
+        }
+
+        private static void Save()
+        {
+            string text = JsonConvert.SerializeObject(Cfg, Formatting.Indented);
+            File.WriteAllText(ConfigPath, text);
+        }
+
+        private static object Get(string field, object def = null)
```
PrimeSonic: This `Get` method could be merged with the `Get<T>` method if the try-catch blocks are moved into the `Save()` and `Load()` methods directly.

Also, you don't need to put `Cfg.TryGetValue` inside of a `try` block. It is perfectly safe to call directly.

QModManager/HarmonyPatches/EnableConsoleSetting.cs
```diff
@@ -0,0 +1,59 @@
...
+    internal static class GraphicsDebugGUI_OnGUI_Patch
+    {
+        // This patch prevents the F3 debug menu from toggling the console
+
+        [HarmonyTranspiler]
+        internal static IEnumerable<CodeInstruction> Transpiler(IEnumerable<CodeInstruction> instructions)
+        {
+            foreach (CodeInstruction instruction in instructions)
+            {
+                if (instruction.opcode == OpCodes.Stsfld && (FieldInfo)instruction.operand == AccessTools.Field(typeof(DevConsole), nameof(DevConsole.disableConsole)))
```
PrimeSonic: Create a temporary variable above the `foreach` so that this only needs to be calculated once
`AccessTools.Field(typeof(DevConsole), nameof(DevConsole.disableConsole)`

Also, you don't need to cast `instruction.operand` for this evaluation.

AlexejheroYTB: image

AlexejheroYTB:
To do: Re-enable the F3 console setting and link it to the qmods one

PrimeSonic:
@AlexejheroYTB please merge this when you get the chance.

We have enough open Pull Requests as it is without having to hold onto ones that have already been reviewed and approved.

AlexejheroYTB:
I will make the required changes and merge this pr sometime this weekend. Thanks for reminding me

PR #126, opened 2020-01-09
Merge pull request #126 from SubnauticaModding/dialog

AlexejheroYTB:


QModManager/Checks/VersionCheck.cs
```diff
@@ -66,16 +69,15 @@ internal static void Parse(string versionStr)
...
-                    {
-                        Dialog.Show(
-                            $"There is a newer version of QModManager available: {latestVersion.ToStringParsed()} (current version: {currentVersion.ToStringParsed()})",
-                            Dialog.Button.download, Dialog.Button.close, true);
-                    }
+                    result = latestVersion;
+                }
+                else if (latestVersion < currentVersion)
+                {
+                    Logger.Info($"Received latest version from GitHub. We're... ahead? Is this a development version, or did Prime forget to update the version again? 🙄");
```
PrimeSonic: Try not to get too cheeky with the log messages.

QModManager/Patching/ModStatus.cs
```diff
@@ -1,10 +1,11 @@
...
 {
     internal enum ModStatus
     {
-        Success,
-        //CanceledByAuthor,
-        CanceledByUser,
+        Merged = -3,
+        //CanceledByAuthor = -2,
+        CanceledByUser = -1,
+        Success = 0,
```
PrimeSonic: The remaining entries in this list should also be given explicit values

PrimeSonic: The remaining entries in this list should also be given explicit values

QModManager/Checks/VersionCheck.cs
PrimeSonic: Try not to get too cheeky with the log messages.

A simple message like this will do
 `"Development build detected: Local version is ahead of server version."`

QModManager/Patching/Patcher.cs
```diff
@@ -101,22 +111,57 @@ internal static void Patch()
                 var initializer = new Initializer(CurrentlyRunningGame);
                 initializer.InitializeMods(modsToLoad);
 
-                int loadedMods = 0;
-                int erroredMods = 0;
-                foreach (QMod mod in modsToLoad)
+                QMod[] loadedMods = modsToLoad.Where(m => m.IsLoaded).ToArray();
+                QMod[] skippedMods = modsToLoad.Where(m => !m.IsLoaded && m.Status < 0).ToArray();
+                QMod[] erroredMods = modsToLoad.Where(m => !m.IsLoaded && m.Status > 0).ToArray();
```
PrimeSonic: Consider using [List<T>FindAll](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.findall?view=netframework-4.8) instead of Linq.

The `modsToLoad` are already in a List, so this method should be available.

QModManager/Utility/Dialog.cs
```diff
@@ -122,16 +156,19 @@ private static IEnumerator ShowDialogEnumerator(string error, Button leftButton,
...
-                texts.Do(t => t.fontSize = t.fontSize + 2);
+                texts.Do(t => t.fontSize = t.fontSize + 4);
 
-                UnityEngine.Object.Destroy(couroutineHandler);
+                UnityEngine.Object.Destroy(coroutineHandler);
 
                 // Re-open the dialog if it is not closeable
-                if (cannotClose)
-                    Show(error, leftButton, rightButton, blue, cannotClose, .05f);
+                if ((leftButtonClicked && leftButton != Button.Close) || (!leftButtonClicked && rightButton != Button.Close)) Show();
```
PrimeSonic: Please keep the code to execute on a separate line from the if-statement

AlexejheroYTB: It's 7 characters

QModManager/Patching/ModStatus.cs
AlexejheroYTB: Everything below Success will be automatically assigned a value > 0

PrimeSonic: Be consistent

PR #131, opened 2020-01-23
Check types after sorting

PrimeSonic:
- Handles CTD bugs caused by duplicate mods
- Handles issues where a disabled or missing dependency causes TypeLoadExceptions when searching for a mod's patch method


QModManager/Patching/IManifestValidator.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace QModManager.Patching
+{
+    internal interface IManifestValidator
```
AlexejheroYTB: What's the point of dependency injection for `internal` classes?

AlexejheroYTB: Unit testing?

PrimeSonic: Yes, and also making it clear how the class is supposed to be used.

There are internal methods that are accessed directly by unit tests,  
but for normal use, only the interface methods should be getting accessed by external actors.

QModManager/SummaryLogger.cs
```diff
@@ -43,26 +42,37 @@ private static void LogStatus(List<QMod> mods, ModStatus statusToReport, string
...
+                switch (statusToReport)
                 {
-                    if (mod.Dependencies.Count() > 0)
+                    case ModStatus.MissingDependency:
                     {
-                        foreach (string dependency in mod.Dependencies)
+                        if (mod.Dependencies.Length > 0)
                         {
-                            if (!QModServices.Main.ModPresent(dependency))
-                                Console.WriteLine($"- {mod.DisplayName} ({mod.Id}) is missing {dependency}");
```
MrPurple6411: An early change I had done to implement this it had also shown all dependencies instead of just the one that is not working. Alex and I had talked about it and both agreed it should really just show the Dependency that was missing and not all dependencies so that it was easier for us to help people as we would see in plain text exactly which dependent had not loaded.

MrPurple6411: Nice work with the switch though. Wish I had thought of it.

PrimeSonic: By the time we have reached the Summary Logger, there is no longer any way to tell which failed dependency set the Status.  
This is why I added the extra logs further up in `QModFactory`.

To be able to print this as a summary, either this data would need to be saved somewhere within the `QMod` instance itself, or check the dependencies again, or figure out another way to handle summary logging entirely.

I'm open to suggestions.

PR #133, opened 2020-02-22
Sorting preferences fix2

PrimeSonic:
### Fix 1:
Inverts the linking actions taken in the methods `LinkAfterPreferences` and `LinkBeforePreferences`.  
Each of those now (correctly) sets links as _after_ and _before_ respectively.

This is opposite to what it was doing before and resolves sorting issues involving only preferences.  
Many unit tests were actually incorrect and have been fixed along with this update.

### Fix 2:
Mods with missing dependencies are allowed to be added to the loading list.  
This is okay since we check for missing dependencies before actually running any patches.

This should allow out logs to better inform these issues.

### Fix 3:
Some members of the `SortedTreeNode` class were reworked for readability.

### Fix 4: 
Fixed naming issue if Mod.json was missing from 1 or more mod.

### Fix 5: 
Fixed Null Ref on listing missing Dependencies.


QModManager/DataStructures/SortedCollection.cs
```diff
@@ -139,7 +139,7 @@ private void LinkAfterPreferences(List<IdType> orderedPreferences, List<SortedTr
                     {
                         if (!item.IsLinked)
                         {
-                            node.SetBefore(item);
+                            node.SetAfter(item);
```
zorgesho: Idk, maybe it's just me, but when I see such a line I read it like this: set `node` after `item` (same as I read `node.SetNodeAfter(item)`) and not like set some `After` variable for `node` to `item`.

Sorting is now works as intended btw.

PrimeSonic: Pushed some renaming updates that should make things a little more readable

PR #140, opened 2020-04-20
Main menu messages

zorgesho:
Public class for adding critical messages to the main menu. They will stay in the main menu and on loading screen.

It's useful for showing errors from QModManager, mods also can use it to show some important messages.

It uses `ErrorMessage` class internally, but it can be safely used by mods during initialization (before `ErrorMessage` is created), so even if mod failed to initialize it still can add some error message.
<details><summary>main menu screenshot</summary>

main_menu
</details>


AlexejheroYTB:
Could you please show a screenshot of how this looks in-game?

QModManager/Utility/MainMenuMessages.cs
```diff
@@ -0,0 +1,169 @@
...
+        /// <summary> Adds message to the main menu. </summary>
+        /// <param name="msg"> message to add </param>
+        /// <param name="size"> text size </param>
+        /// <param name="color"> text color </param>
+        /// <param name="autoformat"> whether it needed to apply formatting tags to the message or show it as it is </param>
+        public static void Add(string msg, int size = defaultSize, string color = defaultColor, bool autoformat = true)
+        {
+            if (Patches.hInstance == null) // just in case
+            {
+                Logger.Error($"MainMenuMessages: trying to add message before Harmony initialized ({msg})");
```
AlexejheroYTB, line 33: You can remove the `MainMenuMessages:` prefix, since that is already added by the `Logger` class

QModManager/Utility/MainMenuMessages.cs
```diff
@@ -0,0 +1,169 @@
...
+
+            if (autoformat)
+            {
+                Assembly callingAssembly = ReflectionHelper.CallingAssemblyByStackTrace(true);
+
+                string prefix;
+                if (callingAssembly != Assembly.GetExecutingAssembly())
+                    prefix = callingAssembly.GetName().Name;
+                else
+                    prefix = "QModManager"; // or maybe just remove prefix if it's QMM ?
```
AlexejheroYTB, line 49: No, I think the prefix should be there even if it's QModManager

QModManager/Utility/MainMenuMessages.cs
```diff
@@ -0,0 +1,169 @@
...
+
+    using Harmony;
+    using UnityEngine;
+    using UnityEngine.SceneManagement;
+
+    /// <summary>
+    /// Allows to add critical messages to the main menu.
+    /// Messages will stay in the main menu and on loading screen.
+    /// </summary>
+    public static class MainMenuMessages
```
PrimeSonic, line 16: This is a `public` feature for use by other mods.  
Move it into the `API` folder & namespace (along with all the other public classes).

I would also recommend making this functionality accessible through the [IQModServices](https://github.com/SubnauticaModding/QModManager/blob/master/QModManager/API/IQModServices.cs) interface.  
Just add the necessary methods to the interface and implement them in [QModServices](https://github.com/SubnauticaModding/QModManager/blob/master/QModManager/API/QModServices.cs) class.

PrimeSonic:
Don't forget to add a new page to the wiki for this.  
https://github.com/SubnauticaModding/QModManager/wiki

QModManager/Utility/MainMenuMessages.cs
zorgesho: > "Move it into the API folder" vs "No need to move"

So, to move or not to move? :)
I think, since `MainMenuMessages` have only one public method I can make this class internal and keep it in Utility namespace and add public method to QModServices as a pass-through. 

Am I understand you correctly?

PrimeSonic: Yeah, sorry. I made things confusing there.  
I said two things cause I was implicitly presenting two options.

The main two things to end up with are:
1. Any public `class` should only exist in (or be accessible through) the `API` namespace.
2. QModServices should be able to expose any public functionality, even if that functionality is implemented somewhere else, just to make discoverability easier.

**Option 1**
Yes, you can absolutely make `MainMenuMessages` be `internal` and have `QModServices` call its methods as a pass-through.

**Option 2**
If you would prefer to keep the `MainMenuMessages` class as `public` and have it usable directly (in addition to `QModServices`), then _that_ is when I would say it should be moved into the `API` namespace.

zorgesho: Ok, I'll go with option 1 then

QModManager/Utility/MainMenuMessages.cs
```diff
@@ -0,0 +1,169 @@
...
+        private const string defaultColor = "red";
+
+        private static List<string> messageQueue;
+        private static List<ErrorMessage._Message> messages;
+
+        /// <summary> Adds message to the main menu. </summary>
+        /// <param name="msg"> message to add </param>
+        /// <param name="size"> text size </param>
+        /// <param name="color"> text color </param>
+        /// <param name="autoformat"> whether it needed to apply formatting tags to the message or show it as it is </param>
```
AlexejheroYTB, line 28: For consistency
```suggestion
        /// <summary>Adds an error message to the main menu.</summary>
        /// <param name="msg">The message to add.</param>
        /// <param name="size">The size of the text.</param>
        /// <param name="color">The color of the text.</param>
        /// <param name="autoformat">Whether or not to apply formatting tags to the message, or show it as it is.</param>
```

QModManager/Utility/MainMenuMessages.cs
```diff
@@ -0,0 +1,169 @@
...
+        {
+            if (Patches.hInstance == null) // just in case
+            {
+                Logger.Error($"MainMenuMessages: trying to add message before Harmony initialized ({msg})");
+                return;
+            }
+
+            Init();
+
+            Logger.Debug($"MainMenuMessages: message added {msg}");
```
AlexejheroYTB, line 39:
```suggestion
            Logger.Debug($"Created message: \"{msg}\"");
```

QModManager/Utility/MainMenuMessages.cs
```diff
@@ -0,0 +1,169 @@
...
+
+                list.InsertRange(index, new List<CodeInstruction>
+                {
+                    new CodeInstruction(OpCodes.Ldloc_S, 6),
+                    new CodeInstruction(OpCodes.Call, AccessTools.Method(typeof(Patches), nameof(_getVal)))
+                });
+
+                return list;
+            }
+        }
```
AlexejheroYTB, line 167: All patching in QModManager has been done via attribute patching.
Why not keep it that way?

zorgesho: Good question, maybe I'm wrong here.
Applying that particular transpiler takes about 200ms on my machine. I'm ok with applying postfix to `ErrorMessage.Awake` every time (~2ms), but 200ms is imho a bit much for unnecessary code, even for initialization, so it applies only if you have messages.

AlexejheroYTB: Hm... Okay. Performance is important.

QModManager/API/QModServices.cs
```diff
@@ -116,5 +117,19 @@ public IQMod GetMod(string id)
...
+        /// Adds a critical message to the main menu.
+        /// Message will stay in the main menu and on the loading screen.
+        /// </summary>
+        /// <param name="msg">The message to add.</param>
+        /// <param name="size">The size of the text.</param>
+        /// <param name="color">The color of the text.</param>
+        /// <param name="autoformat">Whether or not to apply formatting tags to the message, or show it as it is.</param>
+        public void AddCriticalMessage(string msg, int size = MainMenuMessages.defaultSize, string color = MainMenuMessages.defaultColor, bool autoformat = true)
+        {
+            var callingMod = GetMod(ReflectionHelper.CallingAssemblyByStackTrace());
```
AlexejheroYTB, line 131: Will this 100% every time be the correct assembly?

QModManager/Utility/MainMenuMessages.cs
```diff
@@ -0,0 +1,167 @@
...
+
+            var message = ErrorMessage.main.GetExistingMessage(msg);
+            messages.Add(message);
+            message.timeEnd += 1e6f;
+            message.entry.rectTransform.sizeDelta = new Vector2(1920f - newOffset.x * 2f, 0f);
+        }
+
+        private static void OnSceneLoaded(Scene scene, LoadSceneMode _)
+        {
+            if (scene.name != "Main")
```
AlexejheroYTB, line 77: Is that the same case in BZ?

QModManager/API/QModServices.cs
zorgesho: You mean will it be the correct mod? That `CallingAssemblyByStackTrace()` ignores QMM's assembly and system assembly, so it should be correct, I think.

QModManager/Utility/MainMenuMessages.cs
zorgesho: Hmm, good question, I don't know :) Is QMM using same codebase for both? I didn't find any separate configurations.

MrPurple6411: Just checked and both scenes are called main though idk how yours works when all of UWE code has it in lower case.

zorgesho: I don't have BZ, but in SN I see "Main" in code and in log. Where did you see lower case?

MrPurple6411: image

MrPurple6411: Looks like both are Main   and somehow main for both games so if it works on one it should on the other.

PR #141, opened 2020-05-01
Dynamic dialog text handling

PrimeSonic:
* Enables dynamic handling of either `TextMeshProUGUI` or `Text` components in the Dialog box
* Does _not_ require adding a reference to the `TextMeshPro` DLL


README.md
```diff
@@ -137,4 +137,6 @@ ___
 
 ### Merged Mods
 
+###### These mods have had their entire functionality or a part of it merged with QModManager.
+
 - [`Enable Achievements`](https://github.com/AlexejheroYTB/Subnautica-Mods/blob/daf31fa169b923c74defa89d3df29d21a7583e36/EnableAchievements/Mod.cs) by [@AlexejheroYTB](https://github.com/AlexejheroYTB/)
```
AlexejheroYTB, line 142: What if we get rid of this for now? There aren't any plans to "merge" other mods.
```suggestion
```

QModManager/Utility/Dialog.cs
```diff
@@ -172,5 +173,62 @@ private IEnumerator ShowDialogEnumerator()
...
+
+            return textComponents;
+        }
+
+        private static void ChangeAllFontSizes(object[] textComponents, float change)
+        {
+            if (fontSizeProperty == null)
+                return;
+
+            for (int i = 1; i < textComponents.Length; i++)
```
AlexejheroYTB, line 224: Add comment to explain why this loop starts at 1.
```suggestion
            // Loop starts at 1 because text 0 is the main dialog text, which shouldn't be changed
            for (int i = 1; i < textComponents.Length; i++)
```

PR #144, opened 2020-05-07
Fixed copy/pasted grammer typo

PrimeSonic:
Correction to the log line displayed for outdated dependencies


AlexejheroYTB:
Small commits like this one may be done directly on dev.

PR #146, opened 2020-05-23
Fix signs loading vanilla bug

K07H:
=> Fixed the sign loading issue by patching the uGUI_SignInput.UpdateScale() method in a postfix method.


K07H:
Added a check to apply fix only if:
=> BelowZero: any CS version
=> Subnautica: CS version < 65271

K07H:
Actually, let me change the code a little bit so it's more optimized.

PR #147, opened 2020-06-11
Added possibility to define "display duration" for menu messages.

K07H:


QModManager/API/IQModServices.cs
```diff
@@ -30,6 +30,7 @@ public interface IQModServices : IQModAPI
         /// <param name="size">The size of the text.</param>
         /// <param name="color">The color of the text.</param>
         /// <param name="autoformat">Whether or not to apply formatting tags to the message, or show it as it is.</param>
-        void AddCriticalMessage(string msg, int size = MainMenuMessages.defaultSize, string color = MainMenuMessages.defaultColor, bool autoformat = true);
+        /// <param name="timeEnd">The number of seconds the message will stay on screen.</param>
+        void AddCriticalMessage(string msg, int size = MainMenuMessages.defaultSize, string color = MainMenuMessages.defaultColor, bool autoformat = true, float timeEnd = 1e6f);
```
PrimeSonic, line 34: @zorgesho  Can you confirm if changing the method signature like this won't break existing mods that are already using it?

K07H: Hmm you think some mods access this QmodManager method using reflection? That would be weird x)

PrimeSonic: It's a `public` method. Mods can access it directly.

K07H: Yes that's why I said it would be weird, and the only reason it could crash is if you call it using reflection based on parameters signature (which is also a sub category of reflection access).
For me I don't think this breaks anything.

zorgesho: > Hmm you think some mods access this QmodManager method using reflection? That would be weird x)
@K07H Well, I'm doing exactly that :) So players don't need to update their QMM just for that

K07H: > 
> 
> > Hmm you think some mods access this QmodManager method using reflection? That would be weird x)
> 
> @K07H Well, I'm doing exactly that :) So players don't need to update their QMM just for that

Well, you sure you are using parameterized reflection access and not a simple GetMethod(methodName, bindingFlags)?

zorgesho: > GetMethod(methodName, bindingFlags)

If there are two methods with the same name and different signatures it'll throw an exception.

K07H: there's only one method

zorgesho: For some reason I thought this is new second method. No, you can't add optional parameters to the public method.

K07H: Ok I will change method signature for backward compat.

K07H: Except if you don't want people to be able to change display time for critical messages?

zorgesho: Well, I don't see possible use cases for that. These messages are supposed to be used in rare cases, when modder need to be absolutely sure that player has read that message. For other cases there is vanilla `ErrorMessage` class.

K07H: The benefit is that there's already the code for color and stuff so that could save that from having to do it manually when you use ErrorMessage (so it's a little time saving addition).

K07H: The benefit is that there's already the code for color and stuff so that could save that from having to do it manually when you use ErrorMessage (so it's a little time saving addition).

zorgesho: Color and stuff is just this line: https://github.com/SubnauticaModding/QModManager/blob/master/QModManager/Utility/MainMenuMessages.cs#L53
And that code is not runs outside of main menu anyway, because
https://github.com/SubnauticaModding/QModManager/blob/master/QModManager/Utility/MainMenuMessages.cs#L43

K07H: Yes that could save from having to know the syntax of the line you pointed (enjoy already setup parameters ready for you to use), only a small time saving addition.

K07H: Ok after having discussed the pros & cons with @zorgesho it is better to not add this change I initially suggested (even if the method provides all the needed parameters they will not be working in every cases so that doesn't make it a good alternative to the standard ErrorMessage class). PR closed.

PR #148, opened 2020-06-14
Updated Logger class.

K07H:
=> Added a public getter to know if Debug logs are enabled.
=> Adjusted Logger so that it can be used by mods.


K07H:
Done with the PR.

PR #150, opened 2020-06-26
JsonUpdate - Update with the latest Newtonsoft.Json.dll

PrimeSonic:


QModManager/Patching/QMod.cs
```diff
@@ -19,40 +19,40 @@ public QMod()
             // Empty public constructor for JSON
         }
 
-        [JsonProperty(Required = Required.Always)]
+        [JsonProperty(nameof(Id))]
```
zorgesho, line 22: Is `JsonProperty` in the new version really don't use field/property name as its name by default?

PrimeSonic: There is no empty constructor for in this version of `[JsonProperty]`, otherwise I would have used it.

zorgesho: Hmm, what version of Json.NET BZ is using? There is an empty constructor for [JsonPropertyAttribute](https://github.com/JamesNK/Newtonsoft.Json/blob/666d9760719e5ec5b2a50046f7dbd6a1267c01c6/Src/Newtonsoft.Json/JsonPropertyAttribute.cs#L210) in sources.

PrimeSonic: This conversation isn't going anywhere since you aren't looking at the DLL that came with Below Zero.

QModManager/Utility/Config.cs
```diff
@@ -62,8 +62,14 @@ private static void Save()
         {
             try
             {
-                string text = JsonConvert.SerializeObject(Cfg, Formatting.Indented);
-                File.WriteAllText(ConfigPath, text);
+                var serializer = new JsonSerializer
+                {
+                    NullValueHandling = NullValueHandling.Ignore
```
AlexejheroYTB, line 67:
```cs
MissingMemberHandling = MissingMemberHandling.Ignore
```
?

PrimeSonic: [MissingMemberHandling](https://www.newtonsoft.com/json/help/html/P_Newtonsoft_Json_JsonSerializer_MissingMemberHandling.htm) is for _deserialization_.  
It has no impact here for writing the config.

PR #156, opened 2020-07-23
CheckRequiredMods Fix

PrimeSonic:
- Corrections for the CheckRequiredMods method of the ManifestValidator
- Applied better naming of local variables
- Added unit tests for this method and scenarios we can cover

### DIFF
--- a/QModManager/Patching/ManifestValidator.cs
+++ b/QModManager/Patching/ManifestValidator.cs
@@ -46,7 +46,7 @@
 
             if (ProhibitedModIDs.TryGetValue(mod.Id, out ModStatus reason))
             {
-                mod.Status = reason; 
+                mod.Status = reason;
                 return;
             }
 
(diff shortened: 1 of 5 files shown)


tobeyStraitjacket:
Might I suggest we also fix `SummaryLogger` as it doesn't always correctly report missing dependencies. Presently, it only checks in `mod.Dependencies` while sometimes the missing dependency is coming from `mod.VersionDependencies` and in this case the SummaryLogger prints a generic "missing a dependency but none are listed in mod.json" message.

Suggested change: https://github.com/SubnauticaModding/QModManager/blob/CheckRequiredModsFix/QModManager/Utility/SummaryLogger.cs#L49-L65
```cs
                    case ModStatus.MissingDependency:
                        if (mod.RequiredMods != null && mod.RequiredMods.Any())
                        {
                            Console.WriteLine($"- {mod.DisplayName} is missing the following dependencies:");
                            foreach (var requiredMod in mod.RequiredMods)
                            {
                                if (!QModServices.Main.ModPresent(requiredMod.Id) || (requiredMod.RequiresMinimumVersion && QModServices.Main.FindModById(requiredMod.Id).ParsedVersion < requiredMod.MinimumVersion))
                                {
                                    Console.Write($"   - [{requiredMod.Id}]{(requiredMod.RequiresMinimumVersion ? $" at version {requiredMod.MinimumVersion} or greater." : string.Empty)}");
                                }
                            }
                        }
                        else if (mod.Dependencies.Any() || mod.VersionDependencies.Any())
                        {
                            Console.WriteLine($"- {mod.DisplayName} is missing the following dependencies:");
                            foreach (string id in mod.Dependencies.Concat(mod.VersionDependencies.Keys).Distinct())
                            {
                                bool requiresMinimumVersion = mod.VersionDependencies.TryGetValue(id, out var versionString);
                                Version parsedVersion = QModServices.Main.FindModById(id)?.ParsedVersion;
                                bool minimumVersionParsed = false;
                                Version minimumVersion = default;
                                if (requiresMinimumVersion)
                                    minimumVersionParsed = Version.TryParse(versionString, out minimumVersion);

                                if (!QModServices.Main.ModPresent(id) || (requiresMinimumVersion && minimumVersionParsed && parsedVersion < minimumVersion))
                                {
                                    Console.WriteLine($"   - [{id}]{(requiresMinimumVersion ? $" at version {versionString} or greater." : string.Empty)}");
                                }
                                else if (requiresMinimumVersion && !minimumVersionParsed)
                                {
                                    Console.WriteLine($"   - [{id}] failed to parse minimum required version.");
                                }
                            }
                        }
                        else
                        {
                            Console.WriteLine($"- {mod.DisplayName} ({mod.Id}) is missing a dependency but none are listed in mod.json, Please check Nexusmods for list of Dependencies.");
                        }
                        break;
```

PR #167, opened 2020-08-05
Added load order logging for LoadBefore / LoadAfter

tobeyStraitjacket:


QModManager/Patching/ManifestValidator.cs
```diff
@@ -145,15 +145,38 @@ public void CheckRequiredMods(QMod mod)
...
+
+                    Logger.Debug(msg);
+                }
+                if (mod.LoadAfterPreferences.Count > 0)
+                {
+                    string msg = $"{mod.Id} should load after: ";
+                    foreach (var id in mod.LoadAfterPreferences)
+                    {
+                        msg += $"{id} ";
+                    }
```
PrimeSonic, line 176: Consider a small refactor of this repeated code.

```CS
// could be private to the scope of the if-block
string GetModList(string msg, IEnumerable<string> list)
{
    foreach (var id in list)
        msg += $"{id} ";

    return msg;
}

if (requiredMods.Count > 0)
{
    string msg = $"{mod.Id} should load before: ";
    Logger.Debug(GetModList(msg, requiredMods.Values);
}
```

QModManager/Patching/ManifestValidator.cs
```diff
@@ -62,10 +62,10 @@ public void ValidateManifest(QMod mod)
...
                     break;
                 default:
-                {
-                    mod.Status = ModStatus.FailedIdentifyingGame;
-                    return;
-                }
+                    {
+                        mod.Status = ModStatus.FailedIdentifyingGame;
+                        return;
+                    }
```
AlexejheroYTB, line 68: Are these brackets supposed to be here after default?

PrimeSonic: It's just a code style preference that's ultimately just about whitespace/indenting and doesn't really affect anything.

Might as well revert this change since it was probably done by accident.

tobeyStraitjacket: Yeah I have a habit of hitting "format" whenever I save... sorry about that.

QModManager/Patching/ManifestValidator.cs
PrimeSonic: Doesn't have to be that exact code.  
Here's an alternative
```CS
string GetModList(IEnumerable<string> list)
{
    string msg = string.Empty;
    foreach (var id in list)
        msg += $"{id} ";

    return msg;
}

if (requiredMods.Count > 0)
{
    Logger.Debug(($"{mod.Id} should load before: " + GetModList(requiredMods.Values);
}
```

QModManager/Patching/ManifestValidator.cs
tobeyStraitjacket: da8345d

QModManager/Patching/ManifestValidator.cs
tobeyStraitjacket: 4518363

QModManager/Patching/ManifestValidator.cs
```diff
@@ -8,7 +8,6 @@
     using System.IO;
     using System.Linq;
     using System.Reflection;
-    using System.Text.RegularExpressions;
```
tobeyStraitjacket, line 11: This wasn't being used so I cut it.

QModManager/Patching/ManifestValidator.cs
PrimeSonic: Assuming this change works and the `<cast>` returns the Mod ID like we want, then this is fine.  
Alternatively, you can just revert the `RequiredMods` list back to its own method or just undo this refactor as I forgot that not all 3 collections were just string IDs.

PR #169, opened 2020-08-05
Split manifest checks

PrimeSonic:
- Splits the `ManifestValidator` `ValidateManifest` method into separate methods with more focused tasks
- This allows us to add all mods into the app domain in one go without needing to do recursion


PrimeSonic:
Closing. We are keeping the original approach and making improvements to `SortedCollection` instead

PR #170, opened 2020-08-05
Split file tree logs

PrimeSonic:
- Splits the file tree logging into a separate log entry per line
- Uses `Console.Write` directly instead of `Logger` as we do not need to prepend all lines with meta-data
- Solves the issue of sending in one massive log that is too big for the log buffer

### DIFF
--- a/QModManager/Patching/Patcher.cs
+++ b/QModManager/Patching/Patcher.cs
@@ -65,7 +65,8 @@ namespace QModManager.Patching
 
                 try
                 {
-                    Logger.Info($"Folder structure:{IOUtilities.GetFolderStructureAsTree()}");
+                    Logger.Info($"Folder structure:");
+                    IOUtilities.LogFolderStructureAsTree();
                 }
                 catch (Exception e)
                 {
--- a/QModManager/Patching/Patcher.cs
+++ b/QModManager/Patching/Patcher.cs
@@ -65,7 +65,7 @@ namespace QModManager.Patching
 
                 try
                 {
-                    Logger.Info($"Folder structure:");
+                    Logger.Info("Folder structure:");
                     IOUtilities.LogFolderStructureAsTree();
                 }
                 catch (Exception e)
--- a/QModManager/Utility/IOUtilities.cs
+++ b/QModManager/Utility/IOUtilities.cs
@@ -22,13 +22,13 @@ namespace QModManager.Utility
             "steam_shader_cache",
         };
 
-        internal static string GetFolderStructureAsTree(string directory = null)
+        internal static void LogFolderStructureAsTree(string directory = null)
         {
             try
             {
                 directory ??= Environment.CurrentDirectory;
 
-                return GenerateFolderStructure(directory);
+                GenerateFolderStructure(directory);
             }
             catch (Exception e)
             {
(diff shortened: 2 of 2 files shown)


QModManager/Patching/Patcher.cs
```diff
@@ -65,7 +65,8 @@ internal static void Patch()
 
                 try
                 {
-                    Logger.Info($"Folder structure:{IOUtilities.GetFolderStructureAsTree()}");
+                    Logger.Info($"Folder structure:");
```
tobeyStraitjacket, line 68: The string interpolation is not needed here, though I'm just being fussy as I'm not sure if it's gonna waste much performance, probably the compiler is smart enough to just ignore it.

```suggestion
                    Logger.Info("Folder structure:");
```

QModManager/Utility/IOUtilities.cs
```diff
@@ -22,31 +22,36 @@ internal static class IOUtilities
...
         {
             var builder = new StringBuilder();
+
             try
             {
                 builder.AppendLine();
                 builder.AppendLine($"+ {new DirectoryInfo(directory).Name}");
 
+                Logger.Info(builder.ToString());
+                builder.Clear();
```
tobeyStraitjacket, line 49: Suggestion for your consideration:

Instead of modifying the original method in a way that seems a little hacky (clearing the builder rather than just making a new string when you're not using any of the other features of the builder anyway?), why not just take the result from `string GenerateFolderStructure` as usual and split it to an array by `Environment.NewLine`, and loop through the array to print it?

That, or, if you're going to change this method so that is *only* for *printing* the folder structure rather than for *generating and returning it as a string*, I would rename it to `PrintFolderStructure` instead of just removing the `return`.

If you go with the second option (which is viable), I'd essentially recommend ditching the `StringBuilder` entirely and reworking it to just print each line.

QModManager/Utility/IOUtilities.cs
```diff
@@ -22,31 +22,36 @@ internal static class IOUtilities
...
                 foreach (string dir in Directory.GetDirectories(directory))
                 {
                     GetFolderStructureRecursively(builder, dir, 0);
+                    Logger.Info(builder.ToString());
```
tobeyStraitjacket, line 53: Also: Using `Logger.Info` here will result an additional change in the output: each line of the tree will be prefixed with "`[QModManager:INFO] `" which IMO is just added noise and a waste of bytes. I would try to use `Console.WriteLine` instead, except for the initial string "`Folder structure:`".

QModManager/Utility/IOUtilities.cs
PrimeSonic: The intent was to take the contents of each, top-level folder and print that as a single log.  
That would still include several line breaks within the log.

Thoughts?

QModManager/Utility/IOUtilities.cs
PrimeSonic: By convention, we always want to know where a log is coming from.  
Though for this, I can see how it might actually make the log more messy than it has to be.
:+1:

tobeyStraitjacket: > By convention, we always want to know where a log is coming from.
> Though for this, I can see how it might actually make the log more messy than it has to be.
> 👍

While I do agree with that in general, there is a precedent already set in QMM where we are not prefixing each line, but only the first, such as the `SummaryLogger` class. Using `Console.WriteLine()` here for each line *after* the initial has the same effect as `builder.AppendLine()` in the output.

QModManager/Utility/IOUtilities.cs
tobeyStraitjacket: I see! My apologies I somehow completely missed that in your description.

I like the idea. I would still move toward renaming the method as it is now just a printer rather than generating and returning something.

QModManager/Utility/IOUtilities.cs
tobeyStraitjacket: > The intent was to take the contents of each, top-level folder and print that as a single log.
> That would still include several line breaks within the log.

Since I misinterpreted your original intent, I'm not sure my comment here is still valid. I would need to see some sample output before I can make a conclusion :)

PrimeSonic: No, I agree that making an exception in this instance, and keeping the final output in the file formatted the same way as it was before, would be the better choice.  
Do do that, we would have to bypass the `Logger` and go straight through `Console.WriteLine` to avoid prepending the extra meta-data on each log.

PR #174, opened 2020-08-06
Fix logging required mods

tobeyStraitjacket:
Updated the logging of RequiredMods with Debug Logs enabled so as not to throw errors. 

Yes @PrimeSonic, I know it uses LINQ, but it also means its less code-reproduction, and its only on load, and only if Debug Logs are enabled... Take the performance cost ;)


PrimeSonic:
This code only runs at startup as we load the mods to be patched in.  
I'm fine with taking a very minor performance hit in the name of cleaner code.

PR #177, opened 2020-08-06
Aggressively add try{}catch blocks to our preloader patchers

tobeyStraitjacket:
Attempting to catch obscure errors in preloader patches to prevent our preloader patches from cancelling QMM load.

### DIFF
--- a/QMMHarmonyShimmer/QMMHarmonyShimmer.cs
+++ b/QMMHarmonyShimmer/QMMHarmonyShimmer.cs
@@ -33,9 +33,18 @@ namespace QModManager.QMMHarmonyShimmer
         [Obsolete("Should not be used!", true)]
         public static void Initialize()
         {
-            ApplyHarmonyPatches();
-            InitAssemblyResolver();
-            ApplyShims();
+            try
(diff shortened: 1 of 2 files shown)


QMMHarmonyShimmer/QMMHarmonyShimmer.cs
```diff
@@ -33,9 +33,18 @@ public static class QMMHarmonyShimmer
...
         public static void Initialize()
         {
-            ApplyHarmonyPatches();
-            InitAssemblyResolver();
-            ApplyShims();
+            try
+            {
+                ApplyHarmonyPatches();
+                InitAssemblyResolver();
+                ApplyShims();
```
PrimeSonic, line 40: Do we have logs that mark when each of these methods begins?

tobeyStraitjacket: For `ApplyShims` yes, the others are very small, and the `Logger.LogFatal(ex.StackTrace)` should make it pretty clear what method the error is originating from.

PR #190, opened 2020-09-16
Fix system path checks

tobeyStraitjacket:
Fixes an issue experienced by a user who had Subnautica installed in the following path: `E:\Windows Steam Library\steamapps\common\Subnautica\`

Caused by slightly overzealous file IO caution


Data/latest-version.txt
```diff
@@ -1 +1 @@
-4.0.2.3
+4.0.3.0
```
AlexejheroYTB, line 1: 4.0.2.4?

QModManager/Patching/Patcher.cs
```diff
@@ -16,7 +16,7 @@ internal static string QModBaseDir
         {
             get
             {
-                if (Environment.CurrentDirectory.Contains("system32") && Environment.CurrentDirectory.Contains("Windows"))
+                if (Environment.CurrentDirectory.ToLowerInvariant().Contains("system32") && Environment.CurrentDirectory.ToLowerInvariant().Contains("windows"))
```
AlexejheroYTB, line 19: This hasn't happened in a while now and the check can be removed without issues.

QModPluginEmulator/QModPluginGenerator.cs
```diff
@@ -150,7 +150,7 @@ private static void SaveQMMAssemblyCache()
 
         private static void ClearBepInExCache()
         {
-            if (BepInExCachePath.Contains("system32") || BepInExCachePath.Contains("Windows"))
+            if (BepInExCachePath.ToLowerInvariant().Contains("system32") && BepInExCachePath.ToLowerInvariant().Contains("windows"))
```
AlexejheroYTB, line 153: Same here ^

Data/latest-version.txt
tobeyStraitjacket: That's fine, there's another patch that isn't released yet that is 4.0.2.3 which is fixing some install issues, so we can probably actually just keep 4.0.2.3

PrimeSonic (approved):
I see no reason not to include this.

@AlexejheroYTB any comments?

AlexejheroYTB:
The checks were basically put in place in QMM1 when the QMods path was sometimes System32 for whatever reason. That no longer happens for another whatever reason, so the checks dont need to be there.

AlexejheroYTB (approved):

PR #191, opened 2020-09-16
QModManager 4.0.2.3

tobeyStraitjacket:
## Changes in this version
- Added automated cleanup of previous QMM/Nitrox installs #187 
- Removed checking `Windows\System32` path as this is no longer needed and in some cases can cause issues #190


PrimeSonic (approved):

Data/latest-version.txt
```diff
@@ -1 +1 @@
-4.0.2.2
+4.0.2.4
```
AlexejheroYTB, line 1: The version was incremented twice

tobeyStraitjacket: c73a18a

PR #192, opened 2020-09-19
Launch From Exe

MrPurple6411:
This change now makes QMM work when launching from the EXE as well as from the launcher.

Changes the Harmony Patch location in QMMLoader from GameInput.Awake to PlatformUtils.PlatformInitAsyc which causes QMM to not start making its changes until after the game has been restarted if it needs to be.

### DIFF
--- a/QMMLoader/QMMLoader.cs
+++ b/QMMLoader/QMMLoader.cs
@@ -44,7 +44,7 @@ namespace QModManager
         }
 
         private static Harmony harmony;
-        private static MethodInfo entryPointTarget = AccessTools.Method(typeof(GameInput), nameof(GameInput.Awake));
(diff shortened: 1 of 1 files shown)


AlexejheroYTB (commented):
We'll need to contact the developers to inform them of this change and confirm that new targeted method is in no danger of being removed.

MrPurple6411:
> We'll need to contact the developers to inform them of this change and confirm that new targeted method is in no danger of being removed.

this method is the basis of the whole multi platform system and if they remove or change it they will have to remake the whole system. as it is they can just add to it for any number of systems they make the game work for.

tobeyStraitjacket (approved):

PrimeSonic (approved):

AlexejheroYTB (approved):

PR #194, opened 2020-09-21
Fixes for MainMenuMessages

zorgesho:
- Fixed an exception that prevented proper deinitialization
- Fixed conflict with HudConfig mod
- Fixed patch for BelowZero (the same transpiler works for both games now)
- Minor refactoring

### DIFF
--- a/QModManager/Utility/MainMenuMessages.cs
+++ b/QModManager/Utility/MainMenuMessages.cs
@@ -21,11 +21,26 @@
 
         private static readonly Vector2 newOffset = new Vector2(20f, 20f);
         private static Vector2 prevOffset;
+        private static Vector2? prevSize;
 
         private static List<string> messageQueue;
         private static List<ErrorMessage._Message> messages;
 
         private static bool inited => messageQueue != null;
+
+        // we'll restore original offset for messages if position is changed by some other mod (e.g. HudConfig)
+        private class CorrectMsgPos: MonoBehaviour
+        {
+            public void Update()
+            {
+                if (ErrorMessage.main.messageCanvas.localPosition == Vector3.zero)
+                    return;
+
+                ErrorMessage.main.offset = prevOffset;
+                Destroy(this);
+            }
+        }
+
 
         /// <summary>Adds an error message to the main menu.</summary>
         /// <param name="msg">The message to add.</param>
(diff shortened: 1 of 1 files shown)


QModManager/Utility/MainMenuMessages.cs
```diff
@@ -121,98 +137,105 @@ static IEnumerator _waitForLoad()
...
-
-                list.InsertRange(index, new List<CodeInstruction>
+                try
                 {
-                    new CodeInstruction(OpCodes.Ldloc_S, 6),
-                    new CodeInstruction(OpCodes.Call, AccessTools.Method(typeof(Patches), nameof(_getVal)))
-                });
-
-                return list;
+                    var code = new CodeMatcher(cins).
```
PrimeSonic, line 217: Not many of us have used `CodeMatcher` before.  
Could you add some more comments explaining how this all works?

zorgesho: Hmm, I'm not sure where to put comments. Basically, it's just an advanced wrapper around searching in the list of instructions in Traverse/LINQ style. It stores current instruction index, so there is no need to create and check variables for indexes and such.
So in this case it searches for specified method from the end, stores index from the operand of the finded instruction and then continues to search from the same position and then inserts couple of instructions.
It's almost the same transpiler as before, but SN and BZ have slightly different versions of that method and local indexes are not the same for that variable.

PrimeSonic: Just some comments explaining what these methods are doing: `MatchBack`, `Advance`, `MatchForward`, `EmitDelegate`

zorgesho: Well, `CodeMatcher` have its own documentation that is accessible in Visual Studio. And as for github, isn't those method names are straightforward enough to get general idea of what is going on?

PrimeSonic: I honestly find the fluent syntax of `CodeMatcher` to be harder to follow that a hand-crafted state machine doing the same thing.

We've gone back and forth enough here. We can continue this on Discord.

PrimeSonic (approved):
I can get behind this with the added comments

PR #195, opened 2020-10-06
Ignore obsolete mods

tobeyStraitjacket:
Added new `ModStatus`: `ModStatus.Obsolete` for mods so outdated they are no longer supported and will be skipped during the mod loading process.

Added 2 mods to the `ProhibitedModIDs` dictionary in `ManifestValidator.cs` with this new `Obsolete` status:
- AssetTools (`AgonyAssetTools`): Known to be breaking mod options menus for many, many mods. Every mod that every used it is basically dead and/or replaced by another mod.
- QMultiMod (`qmultimod.mod`): Supplanted by several other mods, basically never used, causing compatibility issues with mods that have replaced it.

Added appropriate logging when mods with the `Obsolete` status are skipped.

### DIFF
--- a/QModManager/Patching/ManifestValidator.cs
+++ b/QModManager/Patching/ManifestValidator.cs
@@ -18,6 +18,8 @@
             { "QModManager", ModStatus.BannedID },
             { "QModInstaller", ModStatus.BannedID },
             { "EnableAchievements", ModStatus.Merged },
+            { "AgonyAssetTools", ModStatus.Obsolete },
+            { "qmultimod.mod", ModStatus.Obsolete },
         };
 
         public void ValidateBasicManifest(QMod mod)
--- a/QModManager/Patching/ModStatus.cs
+++ b/QModManager/Patching/ModStatus.cs
@@ -2,6 +2,7 @@
 {
     internal enum ModStatus
     {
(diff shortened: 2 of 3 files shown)


PrimeSonic (approved):

QModManager/Patching/ManifestValidator.cs
```diff
@@ -18,6 +18,8 @@ internal class ManifestValidator : IManifestValidator
             { "QModManager", ModStatus.BannedID },
             { "QModInstaller", ModStatus.BannedID },
             { "EnableAchievements", ModStatus.Merged },
+            { "AgonyAssetTools", ModStatus.Obsolete },
```
AlexejheroYTB, line 21: If we find ourselves in the future needing to change this list very often, we can use a system similar to the version check to get the list from github and cache it in a file.

AlexejheroYTB (approved):

QModManager/Patching/ManifestValidator.cs
PrimeSonic: It is my deepest hope that we never get to a point of needing to update this list frequently.

MrPurple6411 (approved):

PR #211, opened 2021-02-15
Update PirateCheck.cs

ItsCinnabar:
Adds logger text for SmartSteamEmu. Currently you only checked for goldberg. They are about equal in popularity.

### DIFF
--- a/QModManager/Checks/PirateCheck.cs
+++ b/QModManager/Checks/PirateCheck.cs
@@ -17,6 +17,7 @@ namespace QModManager.Checks
             "steam_api64.ini",
             "steam_emu.ini",
             "valve.ini",
+            "SmartSteamEmu.ini",
             "Subnautica_Data/Plugins/steam_api64.cdx",
             "Subnautica_Data/Plugins/steam_api64.ini",
             "Subnautica_Data/Plugins/steam_emu.ini",


PrimeSonic (approved):

AlexejheroYTB (approved):

ItsCinnabar:
To be completely honest, SSE does not have to placed into the subnautica folder. It can be placed anywhere and simply point to the sub.exe. My code adjustment covers only people who place SSE into the direct sub folder. You could for instance place it at the desktop and simply adjust a ini to point to the right folder. No dlls are changed, no sub files are changed. Theres no easy way to tell SSE was used that means, as the entire sub folder is identical. Well, this at least covers the ones who drop it into the sub folder which is most.

Metious:
> To be completely honest, SSE does not have to placed into the subnautica folder. It can be placed anywhere and simply point to the sub.exe. My code adjustment covers only people who place SSE into the direct sub folder. You could for instance place it at the desktop and simply adjust a ini to point to the right folder. No dlls are changed, no sub files are changed. Theres no easy way to tell SSE was used that means, as the entire sub folder is identical. Well, this at least covers the ones who drop it into the sub folder which is most.

if it's placed at the root folder by default, think that's probably enough since most people dont really move anything from their default locations.

PR #219, opened 2021-04-25
Build configs & Auto Compile/Zip

MrPurple6411:
Setup Build Configs for the 4 game Branches (Note the SN.STABLE branch will not work until after the big game update).
Added Inno to the repo to allow for automated compiling.
Setup Automated Exe Compile in post build event 
Setup Auto Vortex Zip creation in Post build event


QModManager/HarmonyPatches/EnableConsoleSetting.cs
```diff
@@ -4,16 +4,19 @@
     using QModManager.Utility;
     using UnityEngine;
 
-    [HarmonyPatch(typeof(DevConsole), nameof(DevConsole.Awake))]
-    internal static class DevConsole_Awake_Patch
+    [HarmonyPatch(typeof(DevConsole), nameof(DevConsole.Update))]
```
PrimeSonic, line 7: Is there a reason we need to patch `Update` now instead of `Awake` or `Start`?  
We don't need this patch to run multiple times.

If we must patch the `Update` method, we should at least try to unpatch after its first run.

QModManager/HarmonyPatches/DisableDevErrorReporting.cs
```diff
@@ -3,13 +3,17 @@
...
+    [HarmonyPatch]
     internal static class SentrySdk_Start_Patch
     {
         // This patch destroys any SentrySdk object
         // The SentrySdk class is the class that sends errors logged with Debug.LogError or Debug.LogException to the Subnautica developers
 
         [HarmonyPrefix]
+        [HarmonyPatch(typeof(SentrySdk), nameof(SentrySdk.Start))]
+#if !SUBNAUTICA_STABLE
+        [HarmonyPatch(typeof(SentrySdk), nameof(SentrySdk.TrySendingEvent))]
```
PrimeSonic, line 15: So on all configs _except_ `SUBNAUTICA_STABLE` we are patching _both_ `Start` _and_ `TrySendingEvent` methods?

QModPluginEmulator/QModManager.QModPluginGenerator.csproj
```diff
@@ -43,23 +63,22 @@
       <HintPath>..\Dependencies\BepInEx\BepInEx\core\Mono.Cecil.dll</HintPath>
       <Private>False</Private>
     </Reference>
-    <Reference Include="Newtonsoft.Json">
-      <HintPath>..\Dependencies\Newtonsoft.Json.dll</HintPath>
+    <Reference Include="Newtonsoft.Json, Version=12.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
+      <HintPath>..\packages\Newtonsoft.Json.12.0.1\lib\net45\Newtonsoft.Json.dll</HintPath>
```
PrimeSonic, line 67: Any reason why we are using NuGet to fetch `Newtonsoft.Json` and `Newtonsoft.Json.Bson` on this config?

QModManager/HarmonyPatches/DisableDevErrorReporting.cs
MrPurple6411: Ahh yes adding the second one was because of the patch timing change and I left the original because well... I forgot it

QModManager/HarmonyPatches/EnableConsoleSetting.cs
MrPurple6411: Unfortunately because of the patch time change we no longer start patching things before the awake or start are called. Not 100% sure why I didnt unpatch it though

MrPurple6411: Actually removed the update as I moved the patch time back before that start so it will work again.

QModManager/HarmonyPatches/DisableDevErrorReporting.cs
MrPurple6411: Actually removed the trysend as I moved the patch time back before that start so it will work again.

QModPluginEmulator/QModManager.QModPluginGenerator.csproj
MrPurple6411: Not that I can see. .. Removing them

PrimeSonic (approved):
Approved for open beta testing

PR #227, opened 2021-05-17
Log file cleaning patch

MrPurple6411:
Log file cleaning patch.
Fix a null ref we caused.
Change DateTime to Now instead of Today

### DIFF
--- a/QModPluginEmulator/QModPluginGenerator.cs
+++ b/QModPluginEmulator/QModPluginGenerator.cs
@@ -82,18 +82,36 @@ namespace QModManager
             }
         }
 
-        public static List<string> DirtyLogStrings = new List<string>()
+        public static List<string> DirtyStartStrings = new List<string>()
         {
-            "(Filename", "Resetting cell with", "Replacing cell",
+            "Resetting cell with", "Replacing cell",
             "PerformGarbage", "Fallback handler could not load"
+        };
+
+        public static List<string> DirtyMidStrings = new List<string>()
+        {
+            "\n(Filename", 
         };
 
         private static void LibcHelper_Format_Postfix(ref string __result)
         {
-            foreach(string dirtyString in DirtyLogStrings)
+            foreach(string dirtyString in DirtyStartStrings)
             {
-                if(__result.Contains(dirtyString))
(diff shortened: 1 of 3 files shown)


QModPluginEmulator/QModPluginGenerator.cs
```diff
@@ -70,6 +82,21 @@ public static void Finish()
...
+        {
+            "(Filename", "Resetting cell with", "Replacing cell",
+            "PerformGarbage", "Fallback handler could not load"
+        };
+
+        private static void LibcHelper_Format_Postfix(ref string __result)
+        {
+            foreach(string dirtyString in DirtyLogStrings)
+            {
+                if(__result.Contains(dirtyString))
```
PrimeSonic, line 95: Use [string.StartsWith](https://docs.microsoft.com/en-us/dotnet/api/system.string.startswith?view=net-5.0) instead of `string.Contains` to keep this from getting slower than it has to be.

zorgesho: Or `IndexOf` can be used from the start

MrPurple6411: problem is not all of these are at the start. that stupid filename thing is actually stuck to the end of a prev line...

zorgesho: I mean, you're using 'IndexOf' in the next line anyway, it's a double work.

MrPurple6411: how about this 
83ae57c

zorgesho: Do you know the first rule of optimization? In this case, any string comparisons in memory are nothing compared to writing log message to disk.

PrimeSonic (approved):

QModPluginEmulator/QModPluginGenerator.cs
PrimeSonic: That's a fair assessment given the ratio between "number of log lines" vs "number of frames drawn".

We can call it good here and expand as needed later.

PR #235, opened 2021-05-29
Applying a Regex to trim out the dirty string patterns instead

tobeyStraitjacket:
Identified a bug with the patch to `MirrorInternalLogs`:

When a log coming from a `BepInEx.Logging.ManualLogSource` comes in directly after one of the mid string trims applied in the patch (e.g. "(Filename: xdgodkdgx)", the patch unfortunately appears to trim both the intended line + the line coming from the `ManualLogSource`, and leaves an empty line behind.

Evidence of this occurring is in the [Fast Loading Screen](https://www.nexusmods.com/subnautica/mods/763) mod. The current release version of QMM for SN.STABLE leaves behind a large amount of empty lines after loading into the game, and Fast Loading Screen's logs are absent.

With the modifications I have made to this patch - which use a `Regex` to strip out occurrences of the intended string - results in the intended behaviour: the garbage line is removed, but no other logs are affected. I have attached an SN.STABLE build of QMM installer from this branch for your testing.

[QModManager_4.1.3_Subnautica_Setup.exe.zip](https://github.com/SubnauticaModding/QModManager/files/6565439/QModManager_4.1.3_Subnautica_Setup.exe.zip)

### DIFF
--- a/QModPluginEmulator/QModPluginGenerator.cs
+++ b/QModPluginEmulator/QModPluginGenerator.cs
@@ -12,15 +12,16 @@ using QModManager.API;
 using QModManager.Patching;
 using QModManager.Utility;
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using System.Text.RegularExpressions;
+using QModManager.API.ModLoading;
 using TypeloaderCache = System.Collections.Generic.Dictionary<string, BepInEx.Bootstrap.CachedAssembly<BepInEx.PluginInfo>>;
 using QMMAssemblyCache = System.Collections.Generic.Dictionary<string, long>;
-using QModManager.API.ModLoading;
-using System.Collections;
 
 namespace QModManager
 {
@@ -57,14 +58,14 @@ namespace QModManager
             {
                 PluginCache = GetPluginCache();
                 Harmony = new Harmony("QModManager.QModPluginGenerator");
-                foreach(Assembly assembly in AppDomain.CurrentDomain.GetAssemblies())
+                foreach (Assembly assembly in AppDomain.CurrentDomain.GetAssemblies())
                 {
-                    if(assembly?.GetName()?.Name?.Contains("MirrorInternalLogs") ?? false)
+                    if (assembly?.GetName()?.Name?.Contains("MirrorInternalLogs") ?? false)
                     {
                         Type type = AccessTools.TypeByName("MirrorInternalLogs.Util.LibcHelper");
                         var method = type?.GetMethod("Format");
 
-                        if(method != null)
+                        if (method != null)
                             Harmony.Patch(method, postfix: new HarmonyMethod(typeof(QModPluginGenerator), nameof(QModPluginGenerator.LibcHelper_Format_Postfix)));
                         break;
                     }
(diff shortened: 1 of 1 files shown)


QModPluginEmulator/QModPluginGenerator.cs
```diff
@@ -57,14 +58,14 @@ public static void Finish()
             {
                 PluginCache = GetPluginCache();
                 Harmony = new Harmony("QModManager.QModPluginGenerator");
-                foreach(Assembly assembly in AppDomain.CurrentDomain.GetAssemblies())
+                foreach (Assembly assembly in AppDomain.CurrentDomain.GetAssemblies())
```
tobeyStraitjacket, line 61: Didn't realise I'd added these spaces, can trim them out if desired. Auto-formatted when I saved.

QModPluginEmulator/QModPluginGenerator.cs
```diff
@@ -88,30 +89,26 @@ public static void Finish()
             "PerformGarbage", "Fallback handler could not load"
         };
 
-        public static List<string> DirtyMidStrings = new List<string>()
+        public static List<string> DirtyPatterns = new List<string>()
```
tobeyStraitjacket, line 92: Actual change starts here

QModPluginEmulator/QModPluginGenerator.cs
```diff
@@ -88,30 +89,26 @@ public static void Finish()
...
             {
-                int i = __result.IndexOf(dirtyString);
-                if(i >= 0)
-                {
-                    __result = __result.Remove(i);
-                    return;
-                }
+                var regex = new Regex(dirtyPattern);
+                __result = regex.Replace(__result, string.Empty).Trim();
             }
```
tobeyStraitjacket, line 112: And ends here

QModPluginEmulator/QModPluginGenerator.cs
```diff
@@ -88,30 +89,26 @@ public static void Finish()
...
-            foreach(string dirtyString in DirtyMidStrings)
+            foreach (string dirtyPattern in DirtyPatterns)
             {
-                int i = __result.IndexOf(dirtyString);
-                if(i >= 0)
-                {
-                    __result = __result.Remove(i);
-                    return;
-                }
+                var regex = new Regex(dirtyPattern);
```
PrimeSonic, line 110: Strongly recommending against using `new Regex` inside a loop.  
Set up the regex instances you need ahead of time, into `static` members, and using `RegexOptions.Compiled` if possible.

QModPluginEmulator/QModPluginGenerator.cs
PrimeSonic, line 92: Could this be `readonly`?

QModPluginEmulator/QModPluginGenerator.cs
tobeyStraitjacket: Fair point, I'll set this up

QModPluginEmulator/QModPluginGenerator.cs
tobeyStraitjacket: Another fair point. I'll change the other List in this codebase to `readonly` while I'm at it.

tobeyStraitjacket: e2e2a8f

QModPluginEmulator/QModPluginGenerator.cs
tobeyStraitjacket: d4990ba

PrimeSonic (approved):

PR #243, opened 2021-06-17
Use 7z for build automation

### LINKED ISSUE #225: Non-compliant zip download
The zip file provided for the manual install of 4.1.2 on Nexus Mods doesn't follow the zip standard and as such can't be used with all zip applications. In particular, the file can't be properly extracted on Linux (I tried both GNOME Archive Manager and command line `unzip`). Instead I ended up having to use a Windows VM to properly extract it.

The problem is the paths inside the zip file contain backslashes. Per the zip standard:

> 4.4.17.1 The name of the file, with optional relative path. The path stored MUST NOT contain a drive or device letter, or a leading slash. All slashes MUST be forward slashes `/` as opposed to backwards slashes `\` for compatibility with Amiga and UNIX file systems etc. If input came from standard input, there is no file name field.

It appears some zip tools in Windows fails to follow this due to an earlier [bug in .NET Framework's zip implementation](https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/mitigation-ziparchiveentry-fullname-path-separator).

I don't know what tool was used to create that zip file, but:
- The 4.0.2.6 download for Subnautica did not have that problem.
- The 4.0.3.4 download for Below Zero also did not have that problem.
- The 4.1.2 download for Below Zero does.

If you could switch back to an archiver that produces proper zip files for future versions, it would save myself (and probably like 2 other people) a lot of trouble. 😄

tobeyStraitjacket:
## Changes in this PR
- Switch to 7z instead of using PowerShell's `Compress-Archive` cmdlet for automation of .zip file creation.
- Resolves #225

### DIFF
--- a/QModPluginEmulator/QModManager.QModPluginGenerator.csproj
+++ b/QModPluginEmulator/QModManager.QModPluginGenerator.csproj
@@ -110,8 +110,7 @@ xcopy "$(TargetDir)QModManager.OculusNewtonsoftRedirect.dll" "$(SolutionDir)Vort
 xcopy "$(TargetDir)QModInstaller.dll" "$(SolutionDir)VortexBuild\$(Configuration)\BepInEx\plugins\QModManager\"  /I /Q /Y
 xcopy "$(TargetDir)QModInstaller.xml" "$(SolutionDir)VortexBuild\$(Configuration)\BepInEx\plugins\QModManager\"   /I /Q /Y
 
-$(SolutionDir)packages\7-Zip.CommandLine.18.1.0\tools\7za.exe a "$(SolutionDir)VortexBuild\QModManager_$(Configuration).zip" "$(SolutionDir)VortexBuild\$(Configuration)\BepInEx" -r
-$(SolutionDir)packages\7-Zip.CommandLine.18.1.0\tools\7za.exe a "$(SolutionDir)VortexBuild\QModManager_$(Configuration).zip" "$(SolutionDir)VortexBuild\$(Configuration)\doorstop_config.ini" "$(SolutionDir)VortexBuild\$(Configuration)\winhttp.dll"
+$(SolutionDir)packages\7-Zip.CommandLine.18.1.0\tools\7za.exe a "$(SolutionDir)VortexBuild\QModManager_$(Configuration).zip" "$(SolutionDir)VortexBuild\$(Configuration)\*"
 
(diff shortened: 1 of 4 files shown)


QModPluginEmulator/QModManager.QModPluginGenerator.csproj
```diff
@@ -107,8 +110,7 @@ xcopy "$(TargetDir)QModManager.OculusNewtonsoftRedirect.dll" "$(SolutionDir)Vort
 xcopy "$(TargetDir)QModInstaller.dll" "$(SolutionDir)VortexBuild\$(Configuration)\BepInEx\plugins\QModManager\"  /I /Q /Y
 xcopy "$(TargetDir)QModInstaller.xml" "$(SolutionDir)VortexBuild\$(Configuration)\BepInEx\plugins\QModManager\"   /I /Q /Y
 
-powershell Compress-Archive -Path '$(SolutionDir)VortexBuild\$(Configuration)\BepInEx' -DestinationPath '$(SolutionDir)VortexBuild\QModManager_$(Configuration).zip' -Force
-powershell Compress-Archive -LiteralPath '$(SolutionDir)VortexBuild\$(Configuration)\doorstop_config.ini', '$(SolutionDir)VortexBuild\$(Configuration)\winhttp.dll' -DestinationPath '$(SolutionDir)VortexBuild\QModManager_$(Configuration).zip' -Update
+$(SolutionDir)packages\7-Zip.CommandLine.18.1.0\tools\7za.exe a "$(SolutionDir)VortexBuild\QModManager_$(Configuration).zip" "$(SolutionDir)VortexBuild\$(Configuration)\*"
```
zorgesho, line 113: I usually use separate .bat file if the build event is more than a few lines. It can be added to the project and edited the same way as sources, with syntax highlighting and without blocking everything.

tobeyStraitjacket: Feel free.

zorgesho: What happened with "wherever we touch the codebase" and all that? :)

tobeyStraitjacket: I didn't say no, I said feel free to make the change.

tobeyStraitjacket: Btw, I didn't write the build event, I am improving it with my modification. If you would like to *further* improve it with a .bat, you are free to do so, nobody is stopping you. I haven't had the time.

MrPurple6411 (approved):
Works on my system.

PR #251, opened 2021-08-23
Allow mods to check what game is currently running.

Esper89:
I noticed that it isn't currently possible for mods to do this without some hacky workarounds.


PrimeSonic:
Haven't seen a need for this feature, but that could just be my own lack of imagination.

The changes look fine.  
Harmless at worst with potential for some modders who want to do more with using a single assembly for mogding both games.

Metious (changes requested):
include the `CurrentlyRunningGame` property to the `IQModServices` interface as well so people can refer to this by doing `QModServices.Main.CurrentlyRunningGame`.

Metious (approved):

PR #253, opened 2021-08-28
Nitrox compatability

MrPurple6411:
Simple Changes to allow Nitrox to run with mods. Please Note this also requires the changes to Nitrox from https://github.com/SubnauticaNitrox/Nitrox/pull/1572 as they also disable QMM when nitrox runs atm.

Changes fatal error to a warning which allows nitrox to run with QModManager.
Added QModService to allow mods to check if nitrox is running to allow for nitrox specific code to only be ran if it is.
Added checks so mods have to specify they are `"NitroxCompat" : true` in the mod.json and if not they are skipped when nitrox is detected as running.
notepad++_d6VRRinNLf

Changes the popup to explain that Nitrox compatability is HIGHLY Experimental and you should expect bugs.
Subnautica_dUI9OFLWjx


PrimeSonic:
I can get behind this.  
I would recommend this eventually be released as `4.3.0` once the final Dev branch is ready.

Metious (approved):

PR #268, opened 2022-04-01
Fix Nitrox dependency errors

toebeann:
Fixes an issue with Nitrox compatibility where mods that are marked as Nitrox compatible will be loaded even if they depend on mods which are _not_ Nitrox compatible, leading to glitchy behaviour.


Metious (approved):
Very good catch

PR #270, opened 2022-04-09
Unify sn1.exp/bz installs

MrPurple6411:
remove audio fixer from check as it is now its own plugin and this check always fails on thunderstore install
PDA.time now exists on SN1.EXP allowing unification of time check. 
match other #if statements in repo 
Rewrite SpriteManager initialize check to not need preprocessor

### DIFF
--- a/QModManager/BepInex/Plugins/QMMLoader.cs
+++ b/QModManager/BepInex/Plugins/QMMLoader.cs
@@ -84,15 +84,14 @@ namespace QModInstaller.BepInEx.Plugins
         {
             while (result.MoveNext())
             {
-                yield return result;
+                yield return result.Current;
             }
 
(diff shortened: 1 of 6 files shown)


MrPurple6411:
Failure due to [QModManager/OptionsManager.cs](https://github.com/SubnauticaModding/QModManager/pull/270/commits/7911e683d43029c868dc71fec2573e0bf30685ca#diff-5858eb0538798761fce9f4e0062786683d6c981184eb01e8d3e406311cfeb64a)

If this method gets the optional in SN1.EXP then we can start having a single build of QMM between SN1.EXP and BZ/BZ.EXP 
Until then we are stuck with 2 unless someone can figure out a way around this.

PR #282, opened 2022-09-20
Masking Userdata in Logfile + Update BZ Version Number of Game

desperationfighter:
In the New BZ Update Personal Data in form of UserID of Steam/EPIC etc is gonna logged to the Log.
Top prevent leaks this info is masked just for Logging.

### DIFF
--- a/QModManager/HarmonyPatches/MaskTelemetryLogging.cs
+++ b/QModManager/HarmonyPatches/MaskTelemetryLogging.cs
@@ -1,0 +1,189 @@
+﻿namespace QModManager.HarmonyPatches
+{
+    using HarmonyLib;
+    using QModManager.Utility;
+
+    //transpiller
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Reflection.Emit;
+
+#if BELOWZERO
+    [HarmonyPatch(typeof(Telemetry),MethodType.Enumerator)]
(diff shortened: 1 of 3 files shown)


QModManager/HarmonyPatches/MaskTelemetryLogging.cs
```diff
@@ -0,0 +1,189 @@
...
+            {
+                Logger.Log(Logger.Level.Debug, $"{thistranspiler} - Deep Logging after-transpiler:");
+                for (int k = 0; k < codes.Count; k++)
+                {
+                    Logger.Log(Logger.Level.Debug, (string.Format("0x{0:X4}", k) + $" : {codes[k].opcode.ToString()}	{(codes[k].operand != null ? codes[k].operand.ToString() : "")}"));
+                }
+            }
+
+            Logger.Log(Logger.Level.Debug, $"{thistranspiler} - Transpiler end going to return");
+            return codes.AsEnumerable();
```
Metious, line 185: Lists are enumerables, so this line can be changed to:
```suggestion
            return codes;
```

Metious (changes requested):
Remove all of the IL logging and IL stack comments, and instead, just keep the "transpiler found" and "transpiler not found" logs.

Metious (approved):

PR #289, opened 2022-12-18
Fixes for the "Living Large" update.

K07H:
A first draft for the "Living Large" update. Tested successfully with the updated version of SMLHelper & Cyclops Docking Mod.


KingTino104 (changes requested):
The mods tab in options doesn't work. (from my testing)

Edwin1212256756:
Bing chilling

K07H:
> The mods tab in options doesn't work. (from my testing)

Yes it's just a quick first draft, there are probably many more things that are not working. Some work is beeing done privately (which is very sad if you want my opinion, because it used to be an open source community driven project, but that's just how it is now!). Anyway, thank you for the report ;)

